PR #209, opened 2019-04-05
Adding code generated type safe wrapper for built in ops

moloneymb:
This makes it easier to use the built in ops from C#.

The code is properly documented including the descriptions for the parameters.

The optional values are actually optional. 

This is not only useful for end users doing low level Tensorflow but it makes it easier when building the high level APIs like Kears for those end users. 

Let me know if you'd like the code moved to a different location with a different class name.


Oceania2018:
Don't have the genertor?

moloneymb:
The generator needs access to the protobuf data files in Tensorflow source code for the generation which makes it pretty complex to build. The C# generator is similar to the TensorFlowSharp projects code generation. 

This generator is written in F# and is similar to https://github.com/fsprojects/TensorFlow.FSharp/blob/master/setup/OperationCodeGenerationCSharp.fsx . 

It's probably a good idea to have the code generator outside of this project in some other repository and just use the generated code here. It will keep TensorFlowNet.Core cleaner

If you're happy to incorporate the generated code I can get to work on a separate code gen repository which I'll port back to C#. 

The F# code is in the zip.
[OperationCodeGenerationFSharp2.zip] https://github.com/SciSharp/TensorFlow.NET/files/3049670/OperationCodeGenerationFSharp2.zip)

Oceania2018:
Actually we use anonymous object instead of dictionary.
image

Your generated function:
image

moloneymb:
You can use both, the dictionary works just fine and handles optional.

PR #217, opened 2019-04-09
Changes to Cond and Control Flow Context

henon:
Please review my changes before merging. I am not 100% sure about them, but without the changes we get NullRefExceptions.  

Check out CreateOpFromTfOperationTest.TestCond test case to see what is still not right with control flow.


Oceania2018:
No issue found. all examples passed.

PR #222, opened 2019-04-11
Update Logistic Regression

SciEvan:
Add the picture of sigmoid


Oceania2018:
Please place `sigmoid.png` in the `_assert` dir, and use relative path `_assert/sigmoid.png` in the markdown file.

PR #228, opened 2019-04-12
cond test with tensor evaluation

henon:


Oceania2018:
image

Oceania2018:
Can we reuse this `eval` function?

PR #235, opened 2019-04-26
Add object detection example and some other minor fixes

captainst:
Add object detection example
Add byte type support in Session and Tensor
Add pbtxtParser class to utility


Oceania2018:
Thanks for your contributing. Appriciate.

PR #237, opened 2019-04-28
added hasattr() and getattr()

arnavdas88:
Hey guys, I havent added the tests for these functions... and do we really need test for these??? I mean they are for base python functions ported to C# and not Tensorflow specific functions....

### DIFF
--- a/src/TensorFlowNET.Core/Python.cs
+++ b/src/TensorFlowNET.Core/Python.cs
@@ -153,6 +153,34 @@ namespace Tensorflow
             }
             return dictionary;
         }
+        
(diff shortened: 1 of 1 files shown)


henon:
I guess they are trivial enough to go without tests

henon:
on the other hand: this is how trivially simple it is to test them:

```Csharp
    [TestClass]
    public class PythonTests
    {
        [TestMethod]
        public void hasattr_getattr()
        {
             var s="1234";
            Assert.IsTrue(hasattr(s, "Length"));
             Assert.AreEqual(4, getattr(s, "Length"));
        }
}
```

PR #239, opened 2019-05-02
Added get_tensor_by_name method to Graph 

JuanuMusic:


Oceania2018:
Appreciate your contribution. BTW, we could get tensor in this way: `Tensor X = graph.get_operation_by_name("Placeholder");`

PR #241, opened 2019-05-07
changed the class Python to be static

kerryjiang:
So if you want to use the methods which are defined in the class Python, just add the below using line:
`
using static Tensorflow.Python;
`


Oceania2018:
It’s very cool.

PR #246, opened 2019-05-15
reKeras Commit

arnavdas88:
Hopefully this works....


arnavdas88:
BTW, I think, we should name it reKeras, as i is no exactly keras, but is more of an inspiration from keras...  It isn't coded like keras actually... Only the layers and and the high-level coding convention has been kept same for simplicity....

Oceania2018:
image

Is it better to rename to `reKeras_ShapeTest` ?

arnavdas88:
Not just the Tests... but also the api...
_Keras.Layers -> reKeras.Layers_   etc.... ***_WDYT_*** ???
And about the redundent force casting.. Yeah... I will remove it **_in the next PR_**...

PR #301, opened 2019-07-22
removing unused namespaces

MovGP0:
removing unused namespaces in order to speed up the compiler and static code analyzers


MovGP0:
tests are also failing in original master branch. errors not introduced here.

Oceania2018:
Johann, are you sure it will speed up the compiler ?

MovGP0:
> Johann, are you sure it will speed up the compiler ?

yes, but usually you wont't notice except in benchmarking. also the autocomplete is cleaner with fewer imports.

MovGP0:
also implemented explicit and implicit type conversion from scalars to tensors and tensors to scalars.

PR #307, opened 2019-07-24
Testing some gradients functions:

sattisvar:
Here are the functions tested
- Concat
- LGamma
- Tanh
- Slice  /!\  Slice seems not to work causing an error `System.InvalidCastException: Unable to cast object of type 'System.Int64' to type 'System.Int32'.`

### DIFF
--- a/test/TensorFlowNET.UnitTest/gradients_test/GradientsTest.cs
+++ b/test/TensorFlowNET.UnitTest/gradients_test/GradientsTest.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NumSharp;
 using Tensorflow;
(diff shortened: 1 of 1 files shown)


Oceania2018:
```python
import tensorflow as tf

a = tf.tanh(tf.constant([2.0, 3.0], shape = [2, 1 ]))
b = tf.strided_slice(a,
                tf.constant([0 ], tf.int32, [1]),
                tf.constant([1 ], tf.int32, [1]),
                tf.constant([1 ], tf.int32, [1]))
grad = tf.gradients(b, a)
with tf.Session() as sess:
    r = sess.run([grad, b])
    print(r)
```

Result:
```console
2019-07-24 12:03:45.462041: I tensorflow/core/platform/cpu_feature_guard.cc:141] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2
[[array([[1.],
       [0.]], dtype=float32)], array([[0.9640276]], dtype=float32)]
```

PR #339, opened 2019-08-01
fix memory leak due to wrong handle sent to api

sattisvar:
A simple typo that led to a memory leak.


Oceania2018:
@sattisvar Can you run through the Unit Test? `All checks have failed`.

PR #341, opened 2019-08-02
Added features to avoid repeated downloads of running data.

Jiuyong:
Perhaps you can add more code to implement the hash check ingress of the file and intermittently pass on.

### DIFF
--- a/src/TensorFlowHub/Utils.cs
+++ b/src/TensorFlowHub/Utils.cs
@@ -25,13 +25,25 @@ namespace Tensorflow.Hub
             if (!Path.IsPathRooted(dirSaveTo))
                 dirSaveTo = Path.Combine(AppContext.BaseDirectory, dirSaveTo);
 
-            if (!Directory.Exists(dirSaveTo))
-                Directory.CreateDirectory(dirSaveTo);
-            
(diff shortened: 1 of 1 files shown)


src/TensorFlowHub/Utils.cs
```diff
@@ -25,13 +25,25 @@ public static async Task DownloadAsync<TDataSet>(this IModelLoader<TDataSet> mod
...
+            if (File.Exists(fileSaveTo))
             {
-                await wc.DownloadFileTaskAsync(url, Path.Combine(dirSaveTo, fileName));
+                //TODO:maybe you can check file's hashcode and "donglowad.info" to complete file ...
+                Console.WriteLine($"{fileSaveTo} already exists.");
             }
+            else
+            {
+                if (!Directory.Exists(dirSaveTo))
+                    Directory.CreateDirectory(dirSaveTo);
```
Nucs: Directory.CreateDirectory dictates in the documentation: 
> Creates all directories and subdirectories in the specified path unless they already exist.

kerryjiang: Good to know, I always check at first, :)

PR #349, opened 2019-08-12
Reworking `global_variables` calls

acifonelli:
- Adding helper function to retrieve vars collections;
- Reworking `global_variables` and `global_variables_initializer`
- Reworking `global_variables` and `global_variables)initializer` in `tf.variables`


acifonelli:
Instead of changing also the `Layers` class I left the signature untouched.
it is building on my side without issues.

PR #358, opened 2019-08-19
added isinstance function

arnavdas88:
Added python internal function **isinstance()** with all overloads

### DIFF
--- a/src/TensorFlowNET.Core/Python.cs
+++ b/src/TensorFlowNET.Core/Python.cs
@@ -309,6 +309,36 @@ namespace Tensorflow
             }
             return (__object__)((object[] args) => { return "NaN"; });
         }
+        public static IEnumerable TupleToEnumerable(object tuple)
+        {
+            Type t = tuple.GetType();
+            if(t.IsGenericType && (t.FullName.StartsWith("System.Tuple") || t.FullName.StartsWith("System.ValueTuple")))
+            {
+                var flds = t.GetFields();
+                for(int i = 0; i < flds.Length;i++)
+                {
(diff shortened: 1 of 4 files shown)


src/TensorFlowNET.Core/Python.cs
```diff
@@ -309,6 +309,36 @@ public static __object__ getattr(object obj, string key, params Type[] ___parame
             }
             return (__object__)((object[] args) => { return "NaN"; });
         }
+        public static IEnumerable TupleToEnumerable(object tuple)
```
Nucs: We are targeting netstandard2.0, you this shouldn't be done via reflection. It should be done by `System.Runtime.CompilerServices.ITuple`.

src/TensorFlowNET.Core/Python.cs
```diff
@@ -309,6 +309,36 @@ public static __object__ getattr(object obj, string key, params Type[] ___parame
...
+                throw new System.Exception("Expected Tuple.");
+            }
+        }
+        public static bool isinstance(object Item1, Type Item2)
+        {
+            return (Item1.GetType() == Item2);
+        }
+        public static bool isinstance(object Item1, object tuple)
+        {
+            var tup = TupleToEnumerable(tuple);
```
Nucs: Avoid using `IEnumerable` and use `ITuple` in a for loop instead.

Nucs:
It turns out `ITuple` was [not included to netstandard 2.0](https://github.com/dotnet/standard/issues/628) therefore this approach is the only current way.

PR #362, opened 2019-08-21
Performance optimization, refactoring and revamping.

Nucs:
#### Changes:
- Refactored `DisposableObject`.
- Added seperate build directory for `TensorflowNET.Examples.GPU`.
- `_FetchHandler`: Switched to `NPTypeCode` (perf-op).
- `gfile.cs, Walk(...)`: Handle case when directory top doesn't exist.
- `Tensor.Creation`: Perf-opted when creating tensor from `NDArray` of dtype `string`.
- `Graph.cs`: refactor and added docs.
- `Tensor.Explicit.cs`: perf-ops.
- `Tensor.Explicit.cs`: Added guards for explicit casts.
- Cleanup and docs to all `Tensor.cs` files.
- Changed all uses of `System.Convert` to `NumSharp.Utilities.Converts`.
- Added all missing types in `dtypes.cs`.
- Renamed `tensor.Data<T>` to `tensor.ToArray<T>`, added obsolete message.
- Renamed `tensor.Data()` to `tensor.BufferToArray()`, added obsolete message.
- Revamped `Tensor.ToArray<T>()` and `Tensor.BufferToArray()`.
- Made `GraphKeys` to use const string instead allocating strings at every `new GraphKeys()` call.
- `Tensor`: Added explicit cast to string.
- `Tensor.DisposeUnmanagedResources`: Nullify _handle after delete.

@Oceania2018, Please review my changes to ensure I did not break something I did not know of.
All unit tests pass. All examples pass (except for those with broken urls)

### DIFF
--- a/src/TensorFlowNET.Core/APIs/c_api.cs
+++ b/src/TensorFlowNET.Core/APIs/c_api.cs
@@ -59,6 +59,6 @@ namespace Tensorflow
         }
 
         [DllImport(TensorFlowLibName)]
-        public static unsafe extern IntPtr TF_Version();
+        public static extern IntPtr TF_Version();
     }
 }
--- a/src/TensorFlowNET.Core/Binding.Util.cs
+++ b/src/TensorFlowNET.Core/Binding.Util.cs
@@ -308,15 +308,14 @@ namespace Tensorflow
         public static IEnumerable TupleToEnumerable(object tuple)
         {
             Type t = tuple.GetType();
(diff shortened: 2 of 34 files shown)


Oceania2018:
Please remove this reference. We don't need it because we install tensorflow binary through NuGet.

image

image

Nucs:
This memory corruption has been fixed in the PR #366

PR #363, opened 2019-08-21
Attempt to provide meta-packages for TensorFlow.lib

chaami:
Hi, the changes provide
- A PowerShell script is available to download the TensorFlow libraries from Google.
- Two nugets (CPU and GPU) can be packaged to provide hopefully easy, light and helpful nugets to end users.
My hope is that when they specify the runtime, only the relevant necessary nuget is deployed.
This has not been tested enough and the PR is to get early feedback 😉 
Thanks.


Nucs:
In order to keep the main solution clean:
Please move the project directories from `TensorFlow.NET/src/runtime-*` to `TensorFlow.NET/redist/*`
Please remove these projects from the main solution and add them into a new solution file `TensorFlow.NET/TensorFlow.NET.Redist.sln`.

Oceania2018:
@harishsk Can you please also review this PR?

chaami:
Hi, sorry for the delay.
1. It makes a lot sense to create another solution, I will push a commit with the change.
1. 7z is only necessary to use the provided PowerShell utility script to extract the TensorFlow libraries from the archives provided by Google (that the script downloads as well).
So it's a requirement needed just by the TensorFlow.NET maintainers when generating a new version of the nuget

Additional notes: 
- By default the project do not pack the nugets when building, one really has to right click on it and request the packing (to make sure people know what they are doing).
- I have set the package version as 1.0.0, let me know if you'd rather have the version match the version of the underlying TensorFlow library that is being bundled.

Kind regards.

Nucs:
1. Looking forward to it.
2. We can't expect to every user of `TF.NET` to have 7zip installed and making it a requirement for using the library is unacceptable in my opinion. After doing some research I found a solid solution: `7Zip4Powershell` which is explained in more details [here](https://stackoverflow.com/a/46876070/1481186).

Oceania2018:
It is needed only when you want to pack new version.

PR #369, opened 2019-08-30
tf2.0

arnavdas88:


src/TensorFlowNET.Core/tf2.cs
```diff
@@ -0,0 +1,49 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+******************************************************************************/
+
+using System;
+
+namespace Tensorflow
+{
+    public static class tf2
```
Oceania2018: What's this `static class tf2` for? Should it be merged into `public partial class tensorflow { }`?

src/TensorFlowNET.Core/Binding.Util.cs
```diff
@@ -333,5 +333,30 @@ public static bool isinstance(object Item1, object tuple)
...
+                var objProp = objProps_w_IndexParams[i];
+                var objPropiParam = objProp.GetIndexParameters();
+                ret.Add(objProp.Name, objProp.GetValue(obj, new object[]{0}));
+            }
+            for(int i = 0; i < objFields.Length; i++)
+                ret.Add(objFields[i].Name, objFields[i].GetValue(obj));
+            return ret;
+            */
+        }
+        public static long id(object obj)
```
Nucs: Can I have reference to the original method?

arnavdas88: the original method is from core python i guess.... it gives the unique id associated to any variable

src/TensorFlowNET.Core/tf2.cs
arnavdas88: actually tf2 is a class that is from tf2.0 and it cant be merged coz it is a different distinct class....

arnavdas88:
And @Oceania2018 , @Nucs  , Can you guys give me a little insight about https://github.com/tensorflow/tensorflow/blob/25006be096cd4f0242a3b979fb212bbd192127b3/tensorflow/python/module/module.py#L332 , here vars() keyword has been used.... but I don't know what data does it extract. just the variables or the functions too???

Nucs:
image
vars should return `Dictionary<string, object>`, it seems to access `module.__dict__` and return its content ([read more here](https://codesachin.wordpress.com/2016/06/09/the-magic-behind-attribute-access-in-python/)).
C#'s equivalent would be to extract all public fields or properties depending on how you plan to do these in classes that inherit Module

src/TensorFlowNET.Core/Binding.Util.cs
Nucs: I can't find an equivalent in python, can you write here a small code that calls `id`?

Nucs:
@arnavdas88 I'm guessing you'll keep pushing to this PR untill you finish your work right? If so, let us know when you need us by requesting reviews.

arnavdas88:
> @arnavdas88 I'm guessing you'll keep pushing to this PR untill you finish your work right? If so, let us know when you need us by requesting reviews.

Awsome...

PR #378, opened 2019-09-08
Added support for unknown dimension #377

### LINKED ISSUE #377: Using unknown dimension in a TensorShape
Currently `TensorShape` only support int/long dimensions. 
We sometimes need to use an unknown dimension that represents a varying batch size that changes from run to run. This is heavily used in reinforcement-learning algorithms.

If you inspect how the following code was written internally, it should explain how python handles None/null dimension
``` python
from tensorflow import TensorShape
from tensorflow.python.framework import tensor_shape

tensorShape: TensorShape = tensor_shape.as_shape((None, 1, 2))
```

The following code is a real use case example
``` python
with tf.Session().as_default() as sess:
    input = tf.placeholder(tf.float64, shape=[None, 3, 3])
    print(input.shape)
    op = tf.layers.flatten(input)
    print(op.shape)
    op = op[3]
    print(op.shape)
    sess.run(tf.global_variables_initializer())

    input_data = np.arange(5 * 3 * 3).reshape((5, 3, 3))
    ret = sess.run(op, feed_dict={input: input_data})

    print(ret)
```

Outputs:
``` python
(?, 3, 3)
(?, 9)
(9,)
[27. 28. 29. 30. 31. 32. 33. 34. 35.]
```

Nucs:
I've decided to make Binding.None to be null - only seems logical to follow python's logic in this case.

Example for use case:
``` C#
var sess = tf.Session().as_default();

var input = tf.placeholder(tf.float64, shape: new TensorShape(None, 3, 3));
//or: var input = tf.placeholder(tf.float64, shape: new []{None, 3, 3});
var op = input[3];
sess.run(tf.global_variables_initializer());

var input_data = np.arange(5 * 3 * 3).reshape((5, 3, 3));
var ret = sess.run(op, (input, input_data));
```

This fixed #377

### DIFF
--- a/src/TensorFlowNET.Core/Binding.cs
+++ b/src/TensorFlowNET.Core/Binding.cs
@@ -10,7 +10,13 @@ namespace Tensorflow
 
         /// <summary>
         ///     Alias to null, similar to python's None.
+        ///     For TensorShape, please use Unknown
         /// </summary>
         public static readonly object None = null;
+
+        /// <summary>
+        /// Used for TensorShape None
+        /// </summary>
+        public static readonly int Unknown = -1;
     }
 }
--- a/src/TensorFlowNET.Core/Binding.cs
+++ b/src/TensorFlowNET.Core/Binding.cs
@@ -7,5 +7,10 @@ namespace Tensorflow
     public static partial class Binding
     {
         public static tensorflow tf { get; } = New<tensorflow>();
+
+        /// <summary>
+        ///     Alias to null, similar to python's None.
+        /// </summary>
+        public static readonly object None = null;
     }
 }
--- a/src/TensorFlowNET.Core/Tensors/TensorShape.cs
+++ b/src/TensorFlowNET.Core/Tensors/TensorShape.cs
@@ -33,7 +33,23 @@ namespace Tensorflow
         /// <summary>
         ///     Returns the size this shape represents.
         /// </summary>
-        public int size => shape.Size;
+        public int size
+        {
+            get
+            {
+                var dims = shape.Dimensions;
+                var computed = 1;
+                for (int i = 0; i < dims.Length; i++)
+                {
+                    var val = dims[i];
+                    if (val <= 0)
+                        continue;
+                    computed *= val;
+                }
+
+                return computed;
+            }
+        }
 
         public TensorShape(TensorShapeProto proto)
         {
--- a/src/TensorFlowNET.Core/Tensors/TensorShape.cs
+++ b/src/TensorFlowNET.Core/Tensors/TensorShape.cs
@@ -3,6 +3,7 @@ using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using NumSharp.Utilities;
 
 namespace Tensorflow
 {
@@ -62,6 +63,30 @@ namespace Tensorflow
                 case 1: shape = Shape.Vector((int) dims[0]); break;
                 case 2: shape = Shape.Matrix(dims[0], dims[1]); break;
                 default: shape = new Shape(dims); break;
+            }
+        }
+
+        /// <summary>
+        ///     An overload that can accept <see cref="Binding.None"/>.
+        /// </summary>
+        public TensorShape(params object[] dims)
+        {
+            var intdims = new int[dims.Length];
+            for (int i = 0; i < dims.Length; i++)
+            {
+                var val = dims[i];
+                if (val == Binding.None)
+                    intdims[i] = -1;
+                else
+                    intdims[i] = Converts.ToInt32(val);
+            }
+
+            switch (dims.Length)
+            {
+                case 0: shape = new Shape(new int[0]); break;
+                case 1: shape = Shape.Vector((int) intdims[0]); break;
+                case 2: shape = Shape.Matrix(intdims[0], intdims[1]); break;
+                default: shape = new Shape(intdims); break;
             }
         }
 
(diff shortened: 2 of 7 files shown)


src/TensorFlowNET.Core/Binding.cs
```diff
@@ -7,5 +7,10 @@ namespace Tensorflow
     public static partial class Binding
     {
         public static tensorflow tf { get; } = New<tensorflow>();
+
+        /// <summary>
+        ///     Alias to null, similar to python's None.
+        /// </summary>
+        public static readonly object None = null;
```
Oceania2018: Why not define it as `int None = -1` ?

src/TensorFlowNET.Core/Tensors/TensorShape.cs
```diff
@@ -57,10 +57,36 @@ public TensorShape(TensorShapeProto proto)
 
         public TensorShape(params object[] dims)
         {
-            var intdims = new int[dims.Length];
```
Oceania2018: Why define it as `params object[] dims` ? If we define `int None = -1`, we can keep it as `params int[] dims` or `params long[] dims`. and can utilize the implicit conversion, we defined conversion for simplify syntax, for example: `public static implicit operator TensorShape((int, int) dims) => new TensorShape(dims.Item1, dims.Item2);`

src/TensorFlowNET.Core/Tensors/TensorShape.cs
```diff
@@ -33,7 +33,23 @@ public class TensorShape
         /// <summary>
         ///     Returns the size this shape represents.
         /// </summary>
-        public int size => shape.Size;
+        public int size
```
Oceania2018: Don't understand why we need this expanded functions? I thought `shape.Size` already computed the size.

test/TensorFlowNET.UnitTest/layers_test/flatten.cs
```diff
@@ -45,5 +45,14 @@ public void Case4()
             var input = tf.placeholder(TF_DataType.TF_INT32, new TensorShape(3, 4, None, 1, 2));
             sess.run(tf.layers.flatten(input), (input, np.arange(3 * 4 * 3 * 1 * 2).reshape(3, 4, 3, 1, 2))).Should().BeShaped(3, 24);
         }
+
+        [TestMethod]
+        public void Case5()
+        {
+            var sess = tf.Session().as_default();
+
+            var input = tf.placeholder(TF_DataType.TF_INT32, new TensorShape(None, 4, 3, 1, 2));
```
Oceania2018: Can we change to `var input = tf.placeholder(tf.int32, (None, 4, 3, 1, 2));`

src/TensorFlowNET.Core/Tensors/TensorShape.cs
Nucs: When shape contains -1 or 0 this will break. TensorShape.size is not something that exists in TF - I've looked into all of the use cases and it is mostly is size is > 0 or > 1

test/TensorFlowNET.UnitTest/layers_test/flatten.cs
Nucs: You'll have to add 6-12 overloads to support all tuples, feel free to add it yourself.

src/TensorFlowNET.Core/Tensors/TensorShape.cs
Nucs: If you add `public TensorShape(params int[] dims)` the project will not compile due to ambiguous methods.

src/TensorFlowNET.Core/Binding.cs
Nucs: I've decided to make Binding.None to be null - only seems logical to follow python's logic in this case.

None is used in many many cases in TF and only when it is in a shape is when we need it as -1. in all other cases it is supposed to be null.

src/TensorFlowNET.Core/Tensors/TensorShape.cs
Nucs: `public static implicit operator TensorShape((int, int) dims) => new TensorShape(dims.Item1, dims.Item2);`
I'll add support if int? and long? in the constructor so you can write:
`public static implicit operator TensorShape((int?, int?) dims) => new TensorShape(dims.Item1, dims.Item2);`

src/TensorFlowNET.Core/Tensors/TensorShape.cs
```diff
@@ -71,61 +71,32 @@ public TensorShape(TensorShapeProto proto)
...
+            {
+                case 0: shape = new Shape(new int[0]); break;
+                case 1: shape = Shape.Vector((int)dims[0]); break;
+                case 2: shape = Shape.Matrix(dims[0], dims[1]); break;
+                default: shape = new Shape(dims); break;
+            }
+        }
 
-            if (dims.Length == 1)
+        public TensorShape(int[][] dims)
```
Nucs: You can remove this constructor, it was valid on for `params object[]`

src/TensorFlowNET.Core/Binding.cs
```diff
@@ -10,7 +10,13 @@ public static partial class Binding
...
         /// <summary>
         ///     Alias to null, similar to python's None.
+        ///     For TensorShape, please use Unknown
         /// </summary>
         public static readonly object None = null;
+
+        /// <summary>
+        /// Used for TensorShape None
+        /// </summary>
+        public static readonly int Unknown = -1;
```
Nucs: This is a horrible solution, just make None = -1, users of the library will never guess how to use this nor will try to dig that deep for examples.

PR #392, opened 2019-09-11
Upgraded NumSharp to 0.20.2

Nucs:
@kerryjiang, Is it ok to upgrade NumSharp to 0.20.2, no breaking changes in your project (TFHub) and compiles successfully.


kerryjiang:
I am fine with the part about TensorflowHub

PR #428, opened 2019-10-19
Fixed a bug that over writes the learning rate when sent as a Tensor.

harshithapv:
Fixed a bug that over-writes the learning rate when sent as a Tensor in GradientDescentOptimizer.cs.


codemzs:
@Oceania2018 can we please merge this and get a nuget out? It’s blocking our release! Thanks!

Issue #350, opened 2019-08-12
Implement tf.image.*


Nucs:
As mentioned in https://github.com/SciSharp/NumSharp/issues/289, we are missing fundamental methods regarding images, for example:
![](https://user-images.githubusercontent.com/649919/62819986-fbacaf00-bb65-11e9-8748-50c2b5e62890.png)

For the full list: https://www.tensorflow.org/api_docs/python/tf/image

Oceania2018:
Function for decode_bmp, decode_gif, decode_jpeg, and decode_png.
https://www.tensorflow.org/api_docs/python/tf/io/decode_image

DerMascha:
Also missing crop to bounding box.

PR #447, opened 2019-11-12
Add image functions

DerMascha:
Added the crop_and_resize method from https://www.tensorflow.org/api_docs/python/tf/image/crop_and_resize.


DerMascha:
#350

PR #448, opened 2019-11-14
Fix collections typing

DerMascha:
As i wrote in chat, all the casts to List<Type> are very fragile and have side effects with different variable types.
Fixes embedding/ crop test.
Also some refactoring/ removed out-commented code/ useless if/else statements.


Oceania2018:
@DerMascha I'm so sorry, I have to rollback this PR. I made a mistake to run through all the examples.
When you want to make changes in tf.net, please make sure you can run all the [examples](https://github.com/SciSharp/TensorFlow.NET-Examples).

PR #528, opened 2020-03-30
Implement SafeStatusHandle

sharwell:
This pull request converts use of `TF_Status*` to a proper `SafeHandle`. This is the first of a rather long sequence of changes to improve the overall reliability of TensorFlow.NET. This particular type was relatively easy to convert, so I started with it. I am happy to answer any questions you may have.

⚠️ This pull request is separated into one commit for each distinct operation. It is highly recommended that the review be commit-by-commit, and the merge (if accepted) preserve the commits in this pull request.

* The first commit introduces `SafeTensorflowHandle` with some extension methods.
* The second commit introduces `SafeStatusHandle` and uses it where appropriate.

📝 I already implemented conversions for many additional resources locally. I'm waiting for feedback on this proposal before submitting each of those individually.


test/TensorFlowNET.UnitTest/Eager/CApi.Eager.cs
```diff
@@ -17,25 +17,23 @@ IntPtr TestMatrixTensorHandle()
...
+            using var status = TF_NewStatus();
 
             var op = TFE_NewOp(ctx, "MatMul", status);
             CHECK_EQ(TF_OK, TF_GetCode(status), TF_Message(status));
             TFE_OpAddInput(op, a, status);
             CHECK_EQ(TF_OK, TF_GetCode(status), TF_Message(status));
             TFE_OpAddInput(op, b, status);
             CHECK_EQ(TF_OK, TF_GetCode(status), TF_Message(status));
-            TF_DeleteStatus(status);
             TFE_OpSetAttrType(op, "T", TFE_TensorHandleDataType(a));
```
sharwell: 📝 This one changed the order of operations, but it seemed acceptable for two reasons:

1. It didn't seem to matter which order these two lines appear in, as the second does not use `status`
1. This is test code, so it's not risking a subtle change to shipping code to switch the order here

A few other tests in the same file have the same situation.

src/TensorFlowNET.Core/Status/Status.cs
```diff
@@ -60,19 +72,17 @@ public void Check(bool throwException = false)
...
+                var message = Message;
+                Console.WriteLine(message);
                 if (throwException)
-                    throw new TensorflowException(Message);
+                    throw new TensorflowException(message);
             }
         }
 
-        public static implicit operator IntPtr(Status status)
-            => status._handle;
```
sharwell: 📝 The implicit operators are _extremely_ problematic, and make the type impossible to use reliably. Explicit operators are equally impossible to use correctly, so I'm removing all of these operators as I work my way through the various types of native resources.

src/TensorFlowNET.Core/Status/Status.cs
```diff
@@ -25,26 +26,37 @@ namespace Tensorflow
...
         /// </summary>
-        public string Message => c_api.StringPiece(TF_Message(_handle));
+        public string Message
+        {
+            get
+            {
+                using (Handle.Lease())
+                {
+                    return StringPiece(TF_Message(Handle));
+                }
```
sharwell: The .NET runtime will automatically acquire a lease when the `SafeHandle` is passed to `TF_Message`. However, the return value from `TF_Message` is an `IntPtr` which is also tied to the lifetime of the `SafeStatusHandle`. By wrapping the use of this pointer in a lease, we ensure the `SafeStatusHandle` is not finalized prior to the completion of the call to `StringPiece`.

📝 It's not normally desirable to use `Handle.Lease()` in coding patterns, but for cases where we _can't_ automatically have things work through proper P/Invoke signatures, it allows the code to still be reliable.

src/TensorFlowNET.Core/Status/Status.cs
```diff
@@ -25,26 +26,37 @@ namespace Tensorflow
     /// TF_Status holds error information. It either has an OK code, or
     /// else an error code with an associated error message.
     /// </summary>
-    public class Status : DisposableObject
```
sharwell: 📝 `DisposableObject` follows legacy resource management patterns that are no longer recommended for use. As types are migrated to new patterns, they will no longer use `DisposableObject`, and eventually the type can simply be deleted.

src/TensorFlowNET.Core/Status/Status.cs
Oceania2018: OK, I don't even know this tricky.

Oceania2018:
Please make this test pass. Below is my master branch.

image

I can't even run (memory crashed) in your branch.

image

sharwell:
@Oceania2018 I will make the tests pass locally. Is there anything we can do to help make the tests pass in CI?

sharwell:
@Oceania2018 I'm getting this error:

```
    Test method TensorFlowNET.UnitTest.ConstantTest.OnesConst threw exception: 
    System.EntryPointNotFoundException: Unable to find an entry point named 'TFE_OpReset' in DLL 'tensorflow'.
```

Oceania2018:
I will upload a test TensorFlow binary to NuGet. The customized package added more TensorFlow C functions.

sharwell:
I'm not sure there's much I can do here prior to a proper dependency getting published. Right now the source refers to a non-existent library:

https://github.com/SciSharp/TensorFlow.NET/blob/784b1ebee9bbced565e1b8729e110451c30f8bc4/src/TensorFlowNET.Core/APIs/c_api.cs#L46

Oceania2018:
@sharwell Sorry about this debug code. `master` branch is still not stable. Will let you know once we published stable `0.20.0`.

Oceania2018:
Sync code, should be working.

sharwell:
@Oceania2018 I updated my branch, but GitHub won't allow me to reopen this issue.

sharwell:
@Oceania2018 If you have questions related to how/why I implemented any part of this the way I did, please let me know. I'm happy to discuss this further.

Oceania2018:
@sharwell Can you sync with master branch and fix the compile issue. The unit tests will be failed as expected, it's OK as long as it can be compiled.

image

sharwell:
@Oceania2018 done!

Oceania2018:
@sharwell It still shows conflicting files.

sharwell:
> It still shows conflicting files.

You pushed to master after my update. 😆 

I'll update again...

sharwell:
@Oceania2018 updated again!

Oceania2018:
@sharwell Can you sync with the upstream ?

image

sharwell:
> Can you sync with the upstream ?

Best to just ignore branches in my fork that aren't part of a pull request. I update directly against this repository.

sharwell:
@Oceania2018 If you ever have questions about how to incorporate the patterns seen here with new APIs/code you are integrating with, please feel free to ask (an `@sharwell` mention is a good way)!

PR #533, opened 2020-04-17
Fixed access violation issue happening on tensorflow native, due susp…

tompetk:
…ection that we pass managed memory pointer to tensorflow native and that gets messed up some time by GC.


Oceania2018:
Still doesn't work. Something exception thrown:
image

Oceania2018:
@tompetk Please try the v0.20 version.

PR #559, opened 2020-06-27
Implement SafeContextHandle as a wrapper for TFE_Context

sharwell:


src/TensorFlowNET.Core/Eager/c_api.eager.cs
```diff
@@ -114,13 +114,13 @@ public partial class c_api
...
-        public static extern TFE_Context TFE_NewContext(IntPtr opts, SafeStatusHandle status);
+        public static extern SafeContextHandle TFE_NewContext(IntPtr opts, SafeStatusHandle status);
 
         [DllImport(TensorFlowLibName)]
-        public static extern TFE_Context TFE_ContextStartStep(IntPtr ctx);
+        public static extern void TFE_ContextStartStep(SafeContextHandle ctx);
 
         [DllImport(TensorFlowLibName)]
-        public static extern TFE_Context TFE_ContextEndStep(IntPtr ctx);
+        public static extern void TFE_ContextEndStep(SafeContextHandle ctx);
```
sharwell, line 123: ⚠️ I changed the signature of this P/Invoke definition. I believe the form here is correct but it should be verified.

sharwell:
I'm actually on the Roslyn team, specifically the team that maintains the IDE tools and developer experience for C# and Visual Basic. I've spent a lot of time in the past working on interop layers though, so I worked on this as part of lending a hand to the ML.NET team.

Oceania2018:
Oh I see, you’re the expert on the .NET language and leading me to the right direction.

PR #560, opened 2020-06-27
Implement SafeContextOptionsHandle as a wrapper for TFE_ContextOptions

sharwell:


test/TensorFlowNET.UnitTest/NativeAPI/Eager/CApi.Eager.Execute_MatMul_CPU.cs
```diff
@@ -19,14 +20,18 @@ public unsafe void Execute_MatMul_CPU()
...
+            static SafeContextHandle NewContext(bool async, SafeStatusHandle status)
+            {
+                using var opts = c_api.TFE_NewContextOptions();
+                c_api.TFE_ContextOptionsSetAsync(opts, Convert.ToByte(async));
+                return c_api.TFE_NewContext(opts, status);
+            }
 
             IntPtr t;
-            using (var ctx = TFE_NewContext(opts, status))
+            using (var ctx = NewContext(async, status))
```
sharwell, line 32: ⚠️ This pull request intentionally uses two different patterns for updating test code. The first pattern, seen here, uses a local function for the creation of the context. This is a bigger change (in terms of lines of code), but allows `using` statements to be used for both `opts` and `ctx` without changing the lifetime of `opts`.

❓ Which of the two patterns do you prefer? I will update this pull request to use the same pattern for all cases.

test/TensorFlowNET.UnitTest/NativeAPI/Eager/CApi.Eager.OpGetInputAndOutputLengths.cs
```diff
@@ -18,7 +18,7 @@ public unsafe void OpGetInputAndOutputLengths()
             var opts = TFE_NewContextOptions();
             using var ctx = TFE_NewContext(opts, status);
             CHECK_EQ(TF_OK, TF_GetCode(status), TF_Message(status));
-            TFE_DeleteContextOptions(opts);
+            opts.Dispose();
```
sharwell, line 21: 📝 This is the second pattern. While the diff is smaller, it has a downside of only allowing the use of one `using` statement, which I've placed on `ctx` instead of `opts`.

test/TensorFlowNET.UnitTest/NativeAPI/Eager/CApi.Eager.Execute_MatMul_CPU.cs
Oceania2018: I like local function style. It looks more structrual.

sharwell: Great! I'll get this updated tomorrow, and when I push to this it will automatically reduce the review to only the new part now that #559 is merged.

sharwell: ➡️ Now updated to this form.

PR #561, opened 2020-06-27
Mark delegates with UnmanagedFunctionPointer

sharwell:


src/TensorFlowNET.Core/APIs/c_api.cs
```diff
@@ -50,8 +50,12 @@ public static string StringPiece(IntPtr handle)
             return handle == IntPtr.Zero ? String.Empty : Marshal.PtrToStringAnsi(handle);
         }
 
+        [UnmanagedFunctionPointer(CallingConvention.Winapi)]
         public delegate void Deallocator(IntPtr data, IntPtr size, ref DeallocatorArgs args);
+
+        [UnmanagedFunctionPointer(CallingConvention.Winapi)]
```
sharwell, line 56: ⚠️ This is the _default_ calling convention, which is used when no other calling convention is explicitly specified. However, it might not be the _correct_ calling convention for this library. You should verify the calling convention used for the library, and then we can make this attribute match.

Oceania2018: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.callingconvention?view=netcore-3.1

PR #565, opened 2020-06-29
Implement SafeBufferHandle as a wrapper for TF_Buffer

sharwell:
Marked as a draft to give me another day to discuss the best way to document `Dangerous*` members in the new API

### DIFF
--- a/src/TensorFlowNET.Core/Buffers/Buffer.cs
+++ b/src/TensorFlowNET.Core/Buffers/Buffer.cs
@@ -14,10 +14,11 @@
    limitations under the License.
 ******************************************************************************/
 
+using NumSharp.Backends.Unmanaged;
 using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using NumSharp.Backends.Unmanaged;
+using Tensorflow.Util;
 using static Tensorflow.c_api;
 
 namespace Tensorflow
@@ -25,34 +26,30 @@ namespace Tensorflow
     /// <summary>
     ///     Represents a TF_Buffer that can be passed to Tensorflow.
     /// </summary>
-    public class Buffer : DisposableObject
+    public sealed class Buffer : IDisposable
     {
-        private unsafe TF_Buffer buffer
-        {
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => *bufferptr;
-        }
+        public SafeBufferHandle Handle { get; }
 
-        private unsafe TF_Buffer* bufferptr
-        {
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => (TF_Buffer*) _handle;
-        }
+        /// <remarks>
+        /// <inheritdoc cref="SafeHandleLease" path="/devdoc/usage"/>
+        /// </remarks>
+        private unsafe ref readonly TF_Buffer DangerousBuffer
+            => ref Unsafe.AsRef<TF_Buffer>(Handle.DangerousGetHandle().ToPointer());
 
         /// <summary>
         ///     The memory block representing this buffer.
         /// </summary>
-        /// <remarks>The deallocator is set to null.</remarks>
-        public UnmanagedMemoryBlock<byte> MemoryBlock
+        /// <remarks>
+        /// <para>The deallocator is set to null.</para>
+        ///
+        /// <inheritdoc cref="SafeHandleLease" path="/devdoc/usage"/>
+        /// </remarks>
+        public unsafe UnmanagedMemoryBlock<byte> DangerousMemoryBlock
         {
             get
             {
-                unsafe
-                {
-                    EnsureNotDisposed();
-                    var buff = (TF_Buffer*) _handle;
-                    return new UnmanagedMemoryBlock<byte>((byte*) buff->data.ToPointer(), (long) buff->length);
-                }
+                ref readonly TF_Buffer buffer = ref DangerousBuffer;
+                return new UnmanagedMemoryBlock<byte>((byte*)buffer.data.ToPointer(), (long)buffer.length);
             }
         }
 
@@ -63,25 +60,23 @@ namespace Tensorflow
         {
             get
             {
-                EnsureNotDisposed();
-                return buffer.length;
+                using (Handle.Lease())
+                {
+                    return DangerousBuffer.length;
+                }
             }
         }
 
-        public Buffer() => _handle = TF_NewBuffer();
+        public Buffer()
+            => Handle = TF_NewBuffer();
 
-        public Buffer(IntPtr handle)
-        {
-            if (handle == IntPtr.Zero)
-                throw new ArgumentException("Handle (IntPtr) can't be zero.", nameof(handle));
+        public Buffer(SafeBufferHandle handle)
+            => Handle = handle;
 
-            _handle = handle;
-        }
+        public Buffer(byte[] data)
+            => Handle = _toBuffer(data);
 
-        public Buffer(byte[] data) : this(_toBuffer(data))
-        { }
-
-        private static IntPtr _toBuffer(byte[] data)
+        private static SafeBufferHandle _toBuffer(byte[] data)
         {
             if (data == null)
                 throw new ArgumentNullException(nameof(data));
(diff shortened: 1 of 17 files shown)


src/TensorFlowNET.Core/Buffers/Buffer.cs
```diff
@@ -14,45 +14,44 @@
...
+        public SafeBufferHandle Handle { get; }
 
-        private unsafe TF_Buffer* bufferptr
-        {
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => (TF_Buffer*) _handle;
-        }
+        /// <remarks>
+        /// <para>The result of this method is only valid in the context of a lease on <see cref="Handle"/>. See
+        /// <see cref="SafeHandleExtensions.Lease(SafeHandle)"/>.</para>
```
sharwell, line 35: 💭 I'm still thinking about ways to describe this. There are actually two valid ways to use it:

1. Direct ownership

    ```csharp
    using (var buffer = new Buffer())
    {
      DoSomethingWithMemory(buffer.DangerousBuffer);
    }
    ```

2. Via a lease on the handle


    ```csharp
    using (var bufferLease = buffer.Handle.Lease())
    {
      DoSomethingWithMemory(buffer.DangerousBuffer);
    }
    ```

src/TensorFlowNET.Core/Buffers/SafeBufferHandle.cs
```diff
@@ -0,0 +1,39 @@
...
+******************************************************************************/
+
+using System;
+using Tensorflow.Util;
+
+namespace Tensorflow
+{
+    public sealed class SafeBufferHandle : SafeTensorflowHandle
+    {
+        public SafeBufferHandle()
```
Oceania2018, line 24: Should we make this constructor private? Becasue when user call `new SafeBufferHandle()` it doesn't impact anything without `handle`.

Or is it better that moving `c_api.TF_NewBuffer()` into constructor?

src/TensorFlowNET.Core/Buffers/Buffer.cs
```diff
@@ -93,38 +90,29 @@ private static IntPtr _toBuffer(byte[] data)
...
-                byte[] data = new byte[len];
-                fixed (byte* dst = data)
-                    System.Buffer.MemoryCopy((void*) bufferptr->data, dst, len, len);
-
-                return data;
+                unsafe
+                {
+                    var source = new ReadOnlySpan<byte>(buffer.data.ToPointer(), checked((int)len));
+                    var data = new byte[len];
+                    source.CopyTo(data);
```
Oceania2018, line 109: Doesn't need `fixed` keyword anymore? Is safe for coping data which might be moved by GC.

src/TensorFlowNET.Core/Buffers/SafeBufferHandle.cs
sharwell: > Should we make this constructor private?

Based on `SafeFileHandle`, it looks like we can make this constructor private. Can you file a follow-up issue for this one since there are a few other cases already merged I'd like to verify as soon as the tests are passing again?

src/TensorFlowNET.Core/Buffers/Buffer.cs
sharwell: This is copying directly to `byte[]`, so there is no need to pin `data` to get a `byte*`.

src/TensorFlowNET.Core/Buffers/SafeBufferHandle.cs
sharwell: Actually I just submitted a second PR for all the other cases (#566).

src/TensorFlowNET.Core/Buffers/Buffer.cs
sharwell: ➡️ I went ahead and simplified this one step further.

src/TensorFlowNET.Core/Buffers/SafeBufferHandle.cs
sharwell: ➡️ Updated `SafeBufferHandle` in this pull request, and the remaining constructors in #566.

src/TensorFlowNET.Core/Buffers/Buffer.cs
sharwell: @Oceania2018 How do you feel about this wording?

> The result of this method is only valid when the underlying `Handle` has not been disposed. If the lifetime of the object is unclear, a lease may be used to prevent disposal while the object is in use. See `SafeHandleExtensions.Lease(SafeHandle)`.

src/TensorFlowNET.Core/Buffers/Buffer.cs
```diff
@@ -14,45 +14,44 @@
...
 
         /// <summary>
         ///     The memory block representing this buffer.
         /// </summary>
-        /// <remarks>The deallocator is set to null.</remarks>
-        public UnmanagedMemoryBlock<byte> MemoryBlock
+        /// <remarks>
+        /// <para>The deallocator is set to null.</para>
+        ///
+        /// <inheritdoc cref="SafeHandleLease" path="/devdoc/usage"/>
```
sharwell, line 47: @Oceania2018 I need to know if you prefer this form, or the form above. This case uses the new `inheritdoc` feature in Quick Info for Visual Studio 2019 to automatically include the same language everywhere you reference this (which also means if you change the original text in the `<devdoc>` element of `SafeHandleLease`, it will update everywhere). The advantage is it avoids copying the same text to many locations. The downside is it relies on an IDE feature to show the information to users. I can use whichever approach you prefer.

src/TensorFlowNET.Core/Util/SafeHandleLease.cs
```diff
@@ -23,6 +23,16 @@ namespace Tensorflow.Util
...
     /// <seealso cref="SafeHandleExtensions.Lease"/>
+    /// <devdoc>
+    /// <para>Elements in this section may be referenced by <c>&lt;inheritdoc&gt;</c> elements to provide common
+    /// language in documentation remarks.</para>
+    ///
+    /// <usage>
+    /// <para>The result of this method is only valid when the underlying handle has not been disposed. If the lifetime
+    /// of the object is unclear, a lease may be used to prevent disposal while the object is in use. See
+    /// <see cref="SafeHandleExtensions.Lease(SafeHandle)"/>.</para>
+    /// </usage>
```
sharwell, line 34: 📝 This is the documentation element which is referenced by the `<inheritdoc>` I mentioned.

src/TensorFlowNET.Core/Buffers/Buffer.cs
Oceania2018: Actually, I don't very understand why we need `buffer.Handle.Lease()` for `public ulong Length` ? Do we really need it once there is `SafeBufferHandle`, GC won't collect `Buffer` if there is any reference to it, right?

src/TensorFlowNET.Core/Buffers/Buffer.cs
Oceania2018: I like it `<devdoc>`, I don't even know that before, glad to learn new things.

sharwell: > I don't even know that before, glad to learn new things.

This pull request is the very first time I've done this. 😄

src/TensorFlowNET.Core/Buffers/Buffer.cs
sharwell: Yes, it's necessary.

1. The implementation of `Length` could get inlined into the caller
2. The call to `DangerousBuffer` is the last true use of the `Buffer` type

The problematic sequence would look like this. I'm using manual inlining to show what could happen.

```
var buffer = new Buffer(); // Note no 'using' statement
var tfBuffer = buffer.DangerousBuffer;
// GC could collect Buffer.Handle here, making tfBuffer invalid
var length = tfBuffer.Length; // Crash!
```

sharwell: 📝 Note that `Lease()` does not incur any allocations, and is generally very fast.

src/TensorFlowNET.Core/Buffers/Buffer.cs
sharwell: ➡️ Both cases use the `inheritdoc` form now.

src/TensorFlowNET.Core/Buffers/Buffer.cs
Oceania2018: I havn't test the performance of `Lease()`, because we donsn't use it very much at this moment. Your comment is making sense.

PR #569, opened 2020-07-01
tensor init bug test

pepure:


pepure:
when I create a constant with 2 dimensions，it's right using quick method like "new[,]{}" , but it occur error when I create Tensor higher than 3 dimensions,"new[,]{}" is not fit ,and "new NDArray(new[, ,] {}" is fit .

PR #574, opened 2020-07-05
Suppress active warnings

sharwell:
Suppress warnings currently appearing in the build.

Eventually these should be fixed, but I didn't want to make the determination for each case at this point. If there is any specific diagnostic ID you would like to see _fixed_ (as opposed to suppressed), let me know and I can make that change.


sharwell:
@Oceania2018 Starting with Visual Studio 2019 version 16.8, when you are working inside the IDE, it will provide hints to remove `#pragma` lines that are no longer necessary. This can help you keep the technical debt for the project low over time.

PR #582, opened 2020-07-07
Implement SafeTensorHandleHandle as a wrapper for TFE_TensorHandle

sharwell:
📝 Marking this one as a draft to see if I can improve documentation and usability in a couple places.


src/TensorFlowNET.Core/Eager/c_api.eager.cs
```diff
@@ -65,7 +66,7 @@ public partial class c_api
         /// <param name="status">TF_Status*</param>
         /// <returns></returns>
         [DllImport(TensorFlowLibName)]
-        public static extern int TFE_OpAddInputList(SafeOpHandle op, IntPtr[] inputs, int num_inputs, SafeStatusHandle status);
+        public static extern int TFE_OpAddInputList(SafeOpHandle op, [In, MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(SafeHandleArrayMarshaler))] SafeTensorHandleHandle[] inputs, int num_inputs, SafeStatusHandle status);
```
sharwell, line 69: 📝 `SafeHandleArrayMarshaler` works for passing arrays of safe handles _to_ unmanaged code, but not the other way around. The `[In]` attribute specifies that I only want to send data one way.

src/TensorFlowNET.Core/Util/SafeHandleArrayMarshaler.cs
```diff
@@ -0,0 +1,132 @@
...
+{
+    internal sealed class SafeHandleArrayMarshaler : ICustomMarshaler
+    {
+        private static readonly SafeHandleArrayMarshaler Instance = new SafeHandleArrayMarshaler();
+
+        private SafeHandleArrayMarshaler()
+        {
+        }
+
+#pragma warning disable IDE0060 // Remove unused parameter (method is used implicitly)
```
sharwell, line 31: 📝 I filed https://github.com/dotnet/roslyn/issues/45743 for this false positive

src/TensorFlowNET.Core/Eager/c_api.eager.cs
Oceania2018: Are you saying we can't passing array from unmanged code to .NET ?

sharwell: We can, but not with the automatic marshaling. See how the `TFE_Execute` call is wrapped for an example.

Oceania2018:
```csharp
            var outputs = new IntPtr[num_outputs];
            if (status.ok())
            {
                c_api.TFE_Execute(op, outputs, ref num_outputs, status.Handle);
                status.Check(true);
            }
```
Can the `outputs ` be Safe Handles `TFE_TensorHandle[]` ?

sharwell: Are you referring to code like this?

https://github.com/SciSharp/TensorFlow.NET/blob/c9f268cdcfd1191c9c92ecbdd59741336f3e391a/test/TensorFlowNET.UnitTest/NativeAPI/Eager/CApi.Eager.OpInferMixedTypeInputListAttrs.cs#L45-L50

Oceania2018: https://github.com/SciSharp/TensorFlow.NET/blob/81215df6e6397d5dbb79ecaa50ed07e997851494/src/TensorFlowNET.Core/Eager/EagerRunner.TFE_Execute.cs#L56

Oceania2018: I just saw you have updated the code. So never mind.

Oceania2018:
@sharwell Sam, are you still working on this?

sharwell:
Sorry for the delay. I'll take a look at this again.

Oceania2018:
@sharwell I've found some UnitTests crashed, for example:
image

When I rollback to old code. It run through.
image

Can you take a look at it and find the root cause?

sharwell:
@Oceania2018 I can take a look, but the current master branch is broken again (mismatch between the version of tensorflow.dll used in the repository and the version of the file the code is compiled against).

Oceania2018:
Sam, you can change 
`public const string TensorFlowLibName = @"D:\SciSharp\tensorflow-google\tensorflow.dll";`
to 
`public const string TensorFlowLibName = "tensorflow.dll";`

sharwell:
The tf.memcpy call here:

https://github.com/SciSharp/TensorFlow.NET/blob/29042df3690601ea4648accd8ceeed647d7ad8e7/test/TensorFlowNET.UnitTest/NativeAPI/Eager/CApi.Eager.cs#L20

Causes memory to be overwritten because of the change introduced by 3a220fa52de83bbc165014f1e6325b58f3261f82 here:

https://github.com/SciSharp/TensorFlow.NET/blob/29042df3690601ea4648accd8ceeed647d7ad8e7/src/TensorFlowNET.Core/tensorflow.memory.cs#L56

src/TensorFlowNET.Core/Eager/c_api.eager.cs
```diff
@@ -90,6 +91,35 @@ public partial class c_api
...
+        /// </summary>
+        /// <remarks>
+        /// Upon successful return, the first <paramref name="num_retvals"/> slots in <paramref name="retvals"/> will
+        /// contain handle instances which the caller is responsible for disposing once they are no longer in use.
+        /// </remarks>
+        /// <param name="op"></param>
+        /// <param name="retvals"></param>
+        /// <param name="num_retvals"></param>
+        /// <param name="status"></param>
+        public static void TFE_Execute(SafeOpHandle op, SafeTensorHandleHandle[] retvals, out int num_retvals, SafeStatusHandle status)
```
sharwell, line 106: 📝 This is really the most "unpleasant" part of this change. The difficulty stems from the fact that `retvals` is created by the caller, and there is no trivial way to dispose of all the resulting handles. My recommendation is to keep this form for now, and focus on the higher-level API if you are interested in cleaning this up. I made the mistake of trying to "fix" this too quickly in an interop library I created in the past, and the result turned out to be more complex.

Oceania2018:
@sharwell Sam can you sync with master?

sharwell:
@Oceania2018 done!

Oceania2018:
image

It should pass. see the master branch.

sharwell:
It looks like one of the tests was incorrectly written to expect a return value, when the call didn't have one. I updated the test to align the expectation with actual behavior.

test/TensorFlowNET.UnitTest/NativeAPI/Eager/CApi.Eager.OpInferMixedTypeInputListAttrs.cs
```diff
@@ -44,15 +42,14 @@ static SafeContextHandle NewContext(SafeStatusHandle status)
...
+                int num_retvals;
+                TFE_Execute(assertOp, retvals, out num_retvals, status);
                 EXPECT_EQ(TF_OK, TF_GetCode(status), TF_Message(status));
-            }
+                EXPECT_EQ(1, num_retvals);
 
-            TFE_DeleteTensorHandle(condition);
-            TFE_DeleteTensorHandle(t1);
-            TFE_DeleteTensorHandle(t2);
-            TFE_DeleteTensorHandle(retvals[0]);
```
sharwell, line 55: 😨 I'm scared to think what this line was doing before

sharwell: Ah, it was hitting this:
https://github.com/tensorflow/tensorflow/blob/d05aa8735f35e6c6d43fdec4c63455912e318298/tensorflow/c/eager/c_api.cc#L940

PR #594, opened 2020-08-04
port constant_value_as_shape & some of its deps

carb0n:
This PR (hopefully; I haven't tested the function other than it not throwing errors at build) implements: 

- `tensor_util.constant_value_as_shape`
- `graph._captures`
- `graph.external_captures`
- `graph.internal_captures`

It's my first PR here, and first time working with C# for more than toy purposes, so don't kill me if something here doesn't work like it should. 👍

### DIFF
--- a/src/TensorFlowNET.Core/Graphs/Graph.cs
+++ b/src/TensorFlowNET.Core/Graphs/Graph.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using System.Runtime.InteropServices;
 using static Tensorflow.Binding;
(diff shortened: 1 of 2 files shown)


carb0n:
Had a question regarding this though regarding the line:

    foreach (Tensor capture in tensor.op.graph.internal_captures())
        if (capture.GetType() == typeof(Tensor))
        {
            ...
        }

Isn't this overly complicated? The python code in the Tensorflow repo checks to see if it's a Tensor too, but i saw somewhere that the captures are both Tensors anyway, so this seems redundant, but I'm not acquainted enough with Tensorflow to know for sure.

Esther2013:
@carb0n Please make sure it can be compiled successfully.

carb0n:
Build succeeded for me after that last commit; fixed an include error.

PR #598, opened 2020-08-05
implement math_ops.real, reduce_variance, reduce_std

carb0n:
* implement math_ops.real, reduce_variance, reduce_std

* add dtype.real_dtype()

* add outward-facing api functions

I tested the build locally and it succeeded with no issues.

### DIFF
--- a/src/TensorFlowNET.Core/Operations/math_ops.cs
+++ b/src/TensorFlowNET.Core/Operations/math_ops.cs
@@ -232,18 +232,19 @@ namespace Tensorflow
        
         public static Tensor real(Tensor input, string name = null)
         {
-            using (var name_ = ops.name_scope(name, "Real", new [] {input}))
+            return tf_with(ops.name_scope(name, "Real", new [] {input}), scope =>
             {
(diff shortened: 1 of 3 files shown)


src/TensorFlowNET.Core/Operations/math_ops.cs
```diff
@@ -229,6 +229,22 @@ public static Tensor sqrt(Tensor x, string name = null)
 
         public static Tensor mul_no_nan<Tx, Ty>(Tx x, Ty y, string name = null)
             => gen_math_ops.mul_no_nan(x, y, name: name);
+       
+        public static Tensor real(Tensor input, string name = null)
+        {
+            using (var name_ = ops.name_scope(name, "Real", new [] {input}))
```
Esther2013, line 235: Please use `tf_with()` instead of using for `ops.name_scope()`.

carb0n: Done, for all of the instances of `ops.name_scope()` added originally.

PR #603, opened 2020-08-18
implement more of tf.image

carb0n:
^

Implements more functions from tf.image, as well as math_ops.saturate_cast, round, reduce_std, and (partially) tensordot.

Code might be messy in places, so I'm opening this now instead of waiting for the TODO to be finished so that y'all can take a look and review it during that time.

TODO:

 - `generate_bounding_box_proposals`: calls a gen_op that doesn't exist yet
 - `_rot90_3D`: does some weird stuff (ex: `[1:, :, :, :]`) that I'm still trying to implement in C#
 - `total_variance`: same situation as `_rot90_3D`

### DIFF
--- a/src/TensorFlowNET.Core/APIs/tf.image.cs
+++ b/src/TensorFlowNET.Core/APIs/tf.image.cs
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
    Copyright 2018 The TensorFlow.NET Authors. All Rights Reserved.
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


carb0n:
I'm failing to replicate the build failures that AppVeyor is showing. I'll take a deeper look but so far, nothing.

carb0n:
I found the source of the build error; I forgot to include some function definitions with this PR, so this should build fine later. Unfortunately, some of them go in gen_ops.cs, which I can't edit online because GitHub says its too big... I'll download GitHub Desktop or something and upload them tomorrow, but for now, I'd say its safe to assume that (after these next commits) it'll build.

PR #623, opened 2020-10-05
Add graph importer input mappings

ChrisUnityArto:
This PR adds the functionality to add input mappings, instead of throwing a `NotImplementedException `when input mappings are provided to the importer.


src/TensorFlowNET.Core/Framework/importer.cs
```diff
@@ -121,7 +121,7 @@ private static void _ProcessNewOps(Graph graph)
 
             foreach(var input in input_map)
             {
-                throw new NotImplementedException("_PopulateTFImportGraphDefOptions");
+                c_api.TF_ImportGraphDefOptionsAddInputMapping(options.Handle, input.Key, 0, input.Value._as_tf_output());
```
Esther2013, line 124: Is the `src_index` fixed value 0? It may cause issue.
Should call `_ParseTensorName` to parse:
```csharp
            foreach (var input in input_map)
            {
                var (src_name, src_index) = _ParseTensorName(input.Key);
                c_api.TF_ImportGraphDefOptionsAddInputMapping(options.Handle, src_name, src_index, input.Value._as_tf_output());
            }
```

Esther2013 (changes requested):

src/TensorFlowNET.Core/Framework/importer.cs
ChrisUnityArto: excellent, thank you for this. I updated the PR as you suggest.

Esther2013 (approved):
Looks good.

Oceania2018:
@ChrisUnityArto Thank you for your PR, we've released to [NuGet v0.21.0](https://www.nuget.org/packages/TensorFlow.NET/0.21.0).

PR #662, opened 2020-12-07
Add F# example to readme

nhirschey:
- Added example of eager mode linear regression to the readme
- Added the example data to the C# example
- Moved the links to documentation to the top of 'How to Use', because that made the most sense to me once we have examples in C# and F#.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,14 @@ In comparison to other projects, like for instance [TensorFlowSharp](https://www
 | tf.net 0.15               | x             | x              |               |
 | tf.net 0.14               | x             |                |               |
 
+Read the docs & book [The Definitive Guide to Tensorflow.NET](https://tensorflownet.readthedocs.io/en/latest/FrontCover.html).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +71,13 @@ int training_steps = 1000;
 float learning_rate = 0.01f;
 int display_step = 100;
 
+// Sample data
+train_X = np.array(3.3f, 4.4f, 5.5f, 6.71f, 6.93f, 4.168f, 9.779f, 6.182f, 7.59f, 2.167f,
```
Esther2013, line 75: Shouldn't be `var train_X = np.array()`?

nhirschey: Not sure, I took it from here:     https://github.com/SciSharp/SciSharp-Stack-Examples/blob/bc4eea022a3d047f0e96e994713ea398b286947c/src/TensorFlowNET.Examples/BasicModels/LinearRegressionEager.cs#L98

nhirschey: The C# notebook example here also does not have `var`:
https://github.com/SciSharp/SciSharpCube/blob/master/home/samples/LinearRegression.ipynb

Esther2013 (approved):

PR #673, opened 2020-12-13
Add test case for tf.transpose

Banyc:


test/TensorFlowNET.UnitTest/Basics/VariableTest.cs
```diff
@@ -83,6 +83,26 @@ public void SliceAssign()
...
+                1, -2, 3
+            };
+            NDArray doubleNd = new double[]
+            {
+                -5, 6, -7
+            };
+            var x = tf.Variable(doubleNd);
+
+            var slice = x[":"];
+            Assert.ThrowsException<System.Exception>(
```
Esther2013, line 100: It's more clear if changing to `ExpectedExceptionAttribute` like below:
```csharp
[TestMethod, Ignore]
[ExpectedException(typeof(System.Exception), "A userId of null was inappropriately allowed.")]
public void NullUserIdInConstructor()
{
   LogonInfo logonInfo = new LogonInfo(null, "P@ss0word");
}
```

PR #675, opened 2020-12-13
Update losses_utils.cs

dataangel:
修改sample_weight初始化

### DIFF
--- a/src/TensorFlowNET.Keras/Utils/losses_utils.cs
+++ b/src/TensorFlowNET.Keras/Utils/losses_utils.cs
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
    Copyright 2018 The TensorFlow.NET Authors. All Rights Reserved.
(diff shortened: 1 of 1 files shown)


src/TensorFlowNET.Keras/Utils/losses_utils.cs
```diff
@@ -25,7 +25,7 @@ public class losses_utils
         public static Tensor compute_weighted_loss(Tensor losses, Tensor sample_weight = null, string reduction = null, string name = null)
         {
             if (sample_weight == null)
-                sample_weight = tf.constant(1.0f);
+                sample_weight = losses.dtype.ToString()=="TF_DOUBLE"? tf.constant(1.0) : tf.constant(1.0f);
```
Oceania2018, line 28: Can you change to `losses.dtyp == TF_DataType.Double` ?

Oceania2018 (changes requested):

PR #740, opened 2021-02-01
Blank SimpleRNN and test for it

MPnoy:
I started implementing RNN, SimpleRNN and related code, but I'm not sure yet that I am writing correctly and not creating a mess. please tell me if there is something here that you definitely don't need to do.


Oceania2018 (approved):
Normally speaking, we'll follow the python version logic.

PR #757, opened 2021-02-19
Adding subset of text preprocessing Keras APIs and unit tests.

NiklasGustafsson:
This adds the Keras Text.Tokenizer API, and adjusts the pad_sequences function to take the output of the Tokenizer without conversion. I tried to stay as close to the Python API as possible, which meant creating a 'Tokenizer()' factory method in the TextApi class.

I added unit tests for the Tokenizer and pad_sequences APIs.

### DIFF
--- a/src/TensorFlowNET.Keras/Preprocessings/Preprocessing.cs
+++ b/src/TensorFlowNET.Keras/Preprocessings/Preprocessing.cs
@@ -10,6 +10,10 @@ namespace Tensorflow.Keras
         public Sequence sequence => new Sequence();
         public DatasetUtils dataset_utils => new DatasetUtils();
 
+        public TextApi text => _text;
+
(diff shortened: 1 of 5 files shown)


NiklasGustafsson:
It's failing on Ubuntu. It seems like all unit tests, not just the ones I added, exit with the same error -- cannot find the 'tensorflow' dependency. Seems to be passing on Windows and MacOS.

Oceania2018:
> It's failing on Ubuntu. It seems like all unit tests, not just the ones I added, exit with the same error -- cannot find the 'tensorflow' dependency. Seems to be passing on Windows and MacOS. Don't know what I've done wrong.

That's OK. no worry.

Esther2013 (approved):
Super cool.

PR #758, opened 2021-02-19
Pooling 1d

NiklasGustafsson:
Added support for some 1D pooling Keras APIs. These can sometimes be useful for processing text with convolutions. Still need to add 1D convolution APIs.

### DIFF
--- a/test/TensorFlowNET.Keras.UnitTest/Layers/PoolingTest.cs
+++ b/test/TensorFlowNET.Keras.UnitTest/Layers/PoolingTest.cs
@@ -1,0 +1,305 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NumSharp;
+using System.Linq;
+using Tensorflow;
+using static Tensorflow.Binding;
+using static Tensorflow.KerasApi;
+
+namespace TensorFlowNET.Keras.UnitTest
+{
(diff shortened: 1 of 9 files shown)


test/TensorFlowNET.Keras.UnitTest/Layers/PoolingTest.cs
```diff
@@ -0,0 +1,305 @@
...
+
+namespace TensorFlowNET.Keras.UnitTest
+{
+    /// <summary>
+    /// https://www.tensorflow.org/versions/r2.3/api_docs/python/tf/keras/layers
+    /// </summary>
+    [TestClass]
+    public class PoolingTest : EagerModeTestBase
+    {
+        private NDArray input_array_1D = np.array(new float[,,]
```
Esther2013, line 16: Normally speaking, we use `np.arange(100).reshape(10, 10)` to generate data.

Esther2013 (approved):

test/TensorFlowNET.Keras.UnitTest/Layers/PoolingTest.cs
NiklasGustafsson: Thanks, that's good to know. I'll take that into account in future test code.

Issue #755, opened 2021-02-19
tf.batch_matmul gives "Attempting to capture an EagerTensor without building a function"


LeeReidCSIRO:
Hi,

I'm getting an exception with `Tensorflow.Binding.tf.batch_matmul` the first time, and only the first time, it is called.

Packages:
SciSharp.TensorFlow.Redist 2.4.1
TensorFlow.Keras 0.4.0
TensorFlow.NET 0.33.0

Consider the following example:

```
    NDArray ta = new NDArray(new float[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }, new Shape(5, 2));
    NDArray tb = new NDArray(new float[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }, new Shape(2, 5));
    Tensor result2 = tf.batch_matmul(ta, tb); // this will throw an exception
```

The exception thrown is:
```
"Attempting to capture an EagerTensor without building a function." | Tensorflow.RuntimeError

stacktrace:
   at Tensorflow.ops.convert_to_tensor(Object value, TF_DataType dtype, String name, Boolean as_ref, TF_DataType preferred_dtype, Context ctx)
   at Tensorflow.OpDefLibrary.<>c__DisplayClass1_0.<_apply_op_helper>b__0(NameScope scope)
   at Tensorflow.Binding.tf_with[TIn,TOut](TIn py, Func`2 action)
   at Tensorflow.OpDefLibrary._apply_op_helper(String op_type_name, String name, Dictionary`2 keywords)
   at Tensorflow.OpDefLibrary._apply_op_helper(String op_type_name, String name, Object args)
   at Tensorflow.gen_math_ops.batch_mat_mul(Tensor x, Tensor y, Boolean adj_x, Boolean adj_y, String name)
   at Tensorflow.tensorflow.batch_matmul(Tensor x, Tensor y)
   at VPG.TF.Test2.LinearRegression2() in C:\somepath\TF\Test2.cs:line 240
   at VPG.Program.Main(String[] args) in C:\somepath\Program.cs:line 14

```
Now, if I run it first in a try-catch statement, the second time it is called it works:

```
    NDArray ta = new NDArray(new float[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }, new Shape(5, 2));
[...]

PR #783, opened 2021-03-19
Fixed many simple functions in gen_math_ops.cs for "Attempting to capture an EagerTensor without building a function"

LeeReidCSIRO:
Hi :)
This is related to Issue #755. 
Multiple functions in gen_math_ops.cs result in "Attempting to capture an EagerTensor without building a function". For example, calling `tf.cos` works, but `tf.acos` results in an exception.

The issue seems to be anything that uses `tf.OpDefLib._apply_op_helper` instead of `tf.Context.ExecuteOp`

This merge fixes the issue for many methods in gen_math_ops. I would suggest having a look at what is left, because presumably these are still buggy.

### DIFF
--- a/src/TensorFlowNET.Core/Operations/gen_math_ops.cs
+++ b/src/TensorFlowNET.Core/Operations/gen_math_ops.cs
@@ -161,18 +161,10 @@ namespace Tensorflow
         }
 
         public static Tensor acos(Tensor x, string name = null)
-        {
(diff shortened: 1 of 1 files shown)


bemobolo:
Hi, tf.reduce_all is doing the same as mentioned in #755. Throws "Attempting to capture..." for the first time, but works when called second time.
TF.NET version 0.70.1

PR #802, opened 2021-04-22
fix test

MPnoy:
Late created GradientTape :/
Now it throws an exception like in the first test


Oceania2018:
It throws exception in python as well.
I pushed the fix https://github.com/SciSharp/TensorFlow.NET/commit/a653914321e23908b6905b6ff2998ee71253ff27
Will close this.

PR #824, opened 2021-07-20
Add Softmax support and throw exception if activation isn't found

Hallupa:


src/TensorFlowNET.Keras/Layers/LayersApi.cs
```diff
@@ -834,7 +835,8 @@ Activation GetActivationByName(string name)
                 "relu" => keras.activations.Relu,
                 "sigmoid" => keras.activations.Sigmoid,
                 "tanh" => keras.activations.Tanh,
-                _ => keras.activations.Linear
```
Oceania2018, line 837: Should we add a default `name` as `linear`?

Hallupa: I think better the user knows the activation wasn't found e.g. if it isn't implemented or is user missspells it.

Before this change I set my activation to 'softmax' but couldn't figure out why the results were wrong. It was only when I looked at the Tensorflow.net source code that I realised softmax wasn't implemented and it was defaulting to linear.

I think the other methods that have defaults should also instead perhaps throw an exception if the value isn't found.

PR #858, opened 2021-09-14
Extending test to use model and demonstrate memory leak

LukeBolly:
Fixes the Session to not immediately dispose the graph.

When running the model there appears to still be a memory leak (though much smaller). Checking the stack after the test has completed shows some tensorflow ops and a graph still in memory:

image


LukeBolly:
@Oceania2018 even if  the memory leak isn't fixed completely, could we get this merged in? It's still a substantial reduction in memory growth compared to the last release with a working SavedModel

Oceania2018:
@LukeBolly Can you resolve the conflicts?

Oceania2018 (approved):

PR #896, opened 2021-12-25
Conv2DTranspose would always throw NotImplementedException regardless…

digital-synapse:
… of dilation_rate used due to a wrong type being used for comparison


Oceania2018 (changes requested):
Can you update the [Equals](https://github.com/SciSharp/TensorFlow.NET/blob/ebfb3275cfc818c7a525dfe28326cf476c2d8526/src/TensorFlowNET.Core/NumPy/ShapeHelper.cs#L89) method as well to fix other potential issues?

It can be like this:
```csharp
case int[] shape3:
    if (shape.ndim != shape3.Length)
        return false;
    return Enumerable.SequenceEqual(shape.as_int_list(), shape3);
```

digital-synapse:
good catch

Oceania2018 (approved):
Appreicate for your contribution.

PR #897, opened 2021-12-25
Fix duplicate graph node name

digital-synapse:
testing a conv2d autoencoder from
https://keras.io/guides/functional_api/#:~:text=keras.,graph%20(DAG)%20of%20layers

revealed that layer names ending in / could be duplicated in the graph, preventing the model from being compiled. This fix resolved the duplicate layer name issue.


Oceania2018 (approved):

Oceania2018 (commented):
I revert this change due to other issue threw.

image

Issue #822, opened 2021-07-14
Gradient of tf.reduce_sum(_, axis = _) throws an exception


mishun:
Hi!

This example (it creates 2x2 matrix, uses tf.reduce_sum on it and tries to differentiate the result):
```F#
open Tensorflow
open type Tensorflow.Binding

[<EntryPoint>]
let main argv =
    tf.compat.v1.disable_eager_execution()
    let x = tf.placeholder(tf.float64, shape = TensorShape(2))
    let sq = tf.matmul(tf.reshape(x, shape = TensorShape(2, 1)), tf.reshape(x, shape = TensorShape(1, 2))) // Make 2z2 matrix
    let y = tf.reduce_sum(sq, axis = 0)
    let g = tf.gradients(y, x)
    begin
        use session = tf.Session()
        let res = session.run(g.[0], [| FeedItem(x, [| -1.0 ; 1.0 |]) |])
        printfn "result = %O" res
    end
    0
```
causes following exception:
```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NotImplementedException
   at Tensorflow.Gradients.math_grad._SumGrad(Operation op, Tensor[] grads)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.RuntimeType.InvokeMember(String name, BindingFlags bindingFlags, Binder binder, Object target, Object[] providedArgs, ParameterModifier[] modifiers, CultureInfo culture, String[] namedParams)
   at Tensorflow.ops.<>c__DisplayClass1_1.<RegisterFromAssembly>b__3(Operation oper, Tensor[] out_grads)
   at Tensorflow.gradients_util._MaybeCompile(String scope, Operation op, Tensor[] out_grads, Action func, Func`3 grad_fn)
   at Tensorflow.gradients_util.<>c__DisplayClass0_2.<_GradientsHelper>b__6(NameScope scope1)
   at Tensorflow.Binding.tf_with[T](T py, Action`1 action)
   at Tensorflow.gradients_util.<>c__DisplayClass0_0.<_GradientsHelper>b__0(NameScope scope)
[...]

PR #917, opened 2022-03-09
Add test case for tf.reduce_sum(..., axis = ...) (issue #822)

mishun:


mishun:
I noticed that CI build hasn't failed after merge. It doesn't run tests?

Oceania2018:
It shows successful, is it supposed to failed?

mishun:
Yes, new test case currently fails with NotImplementedException. From what I can see, CI build doesn't run test from `TensorFlowNET.Graph.UnitTest`.

PR #940, opened 2022-06-19
Fix Conv2dTranspose bias

Fiffers:
Conv2dTranspose defaults to have use_bias = true but currently throws a not implemented exception when the parameter is true.


AsakusaRinne:
@Oceania2018 Shall we merge this PR?

Issue #912, opened 2022-02-17
"Cannot access a disposed object." with multithreading 


Superpiffer:
I have an issue with Tensorflow.NET 0.70.1 and a multithreading console application:
![image](https://user-images.githubusercontent.com/1334453/154437771-e5533571-5a28-48c7-be6d-682da5bc02fc.png)
I have the same problem with CPU or GPU native library, tensorflow 2.7.0.
Here you can find a sample project:
[TensorflowMultithreadingTest.zip](https://github.com/SciSharp/TensorFlow.NET/files/8087008/TensorflowMultithreadingTest.zip)

Note: I cannot use `using` for the inference because I need to reuse the same session, I don't think is feasible to create a new Session every time.
Note 2: Using `modelSession.Dispose()` at the end of the processing or using `GC.SuppressFinalize(modelSession)` after the declaration solve the problem, but I don't think it's the best solution.

Superpiffer:
@Oceania2018 I have an update on this issue:
- I have this problem only on **.NET Framework 4.8**, not on .NET 5 or .NET 6 for some reason. I read online that maybe there are some difference in the finalize phase between .NET Framework and .NET Core: https://stackoverflow.com/a/44735403
- As a workaround, in my project I created a new Session in a wrapper class, in this class I disabled the finalizer with `GC.SuppressFinalize(_session)` and I manually deleted the session in the finalizer of the wrapper using a new Status object and not the static one: 
`~Session() => c_api.TF_DeleteSession(_session, _status.Handle)`

Wrapper class:
```csharp
internal class Session
{
    private readonly Tensorflow.Session _session;
    private readonly Status _status = new Status();
    public Graph graph => _session.graph;

    public Session(Graph graph, ConfigProto config)
    {
        _session = new Tensorflow.Session(graph, config, _status);
        GC.SuppressFinalize(_session);
    }

    public NDArray run(Tensor tensor, FeedItem item) => _session.run(tensor, item);
    public NDArray[] run(Tensor[] tensors, FeedItem item) => _session.run(tensors, item);

    ~Session() => c_api.TF_DeleteSession(_session, _status.Handle);
}
```

I don't know if this make sense, but apparently it works for now...

Oceania2018:
@Superpiffer Is it possible to PR a patch with compile condition?

Superpiffer:
> @Superpiffer Is it possible to PR a patch with compile condition?

[...]

PR #975, opened 2023-02-01
Removed use of tf.Status static instance

Superpiffer:
In multithreading .NET 4.8 applications, sometimes in Session finalizer the method c_api.TF_DeleteSession find f.Status static instance already disposed for some reason. No problem for .NET 6 application or with a single thread. (https://github.com/SciSharp/TensorFlow.NET/issues/912)

### DIFF
--- a/src/TensorFlowNET.Core/Sessions/BaseSession.cs
+++ b/src/TensorFlowNET.Core/Sessions/BaseSession.cs
@@ -291,7 +291,7 @@ namespace Tensorflow
         protected override void DisposeUnmanagedResources(IntPtr handle)
         {
             // c_api.TF_CloseSession(handle, tf.Status.Handle);
-            c_api.TF_DeleteSession(handle, tf.Status.Handle);
+            c_api.TF_DeleteSession(handle, c_api.TF_NewStatus());
         }
     }
 }


Oceania2018:
@Superpiffer Do you think it will work if I change to below
``` csharp
if (tf.Status == null || tf.Status.Handle.IsInvalid)
{
    c_api.TF_DeleteSession(handle, c_api.TF_NewStatus());
}
else
{
    c_api.TF_DeleteSession(handle, tf.Status.Handle);
}
```

Superpiffer:
> @Superpiffer Do you think it will work if I change to below
> 
> ```cs
> if (tf.Status == null || tf.Status.Handle.IsInvalid)
> {
>     c_api.TF_DeleteSession(handle, c_api.TF_NewStatus());
> }
> else
> {
>     c_api.TF_DeleteSession(handle, tf.Status.Handle);
> }
> ```

I think so, maybe that's ok as a "last resort" in my use case, also preserving the previous behaviour. I can test it on monday.
The problem is that I'm not familiar with ThreadLocal and I don't know why tf.Status is disposed before all the session threads are stopped and ONLY on .NET 4.8...

Superpiffer:
@Oceania2018 I tried and I have the same problem:
image
image

Superpiffer:
> Is there any code like `using(session.as_default()){ }` in your code? Sometimes annoying things happens when I use the scope with `using`.

No, I cannot use `using ` because I need to reuse the same session during my processing. You can find a sample code here: https://github.com/SciSharp/TensorFlow.NET/issues/912

Also I made a new PR with a possible solution: https://github.com/SciSharp/TensorFlow.NET/pull/978

PR #989, opened 2023-03-01
Support loading of SavedModel format

AsakusaRinne:
Add loading of SavedModel format.


AsakusaRinne:
Note: `keras.generic_utils.deserialize_keras_object` and `keras.generic_utils.deserialize_layer_args` has many duplicated implementations now due to the time limitation. It needs to be optimized in the future.

src/TensorFlowNET.Keras/Utils/generic_utils.cs
```diff
@@ -51,6 +56,85 @@ public static JObject serialize_keras_object(IKerasConfigable instance)
             return serialize_utils.serialize_keras_class_and_config(instance.GetType().Name, config, instance);
         }
 
+        public static Layer deserialize_keras_object(string class_name, JToken config)
```
Oceania2018, line 59: We can dynamically create an Layer object and add some behavior in `IKerasConfig`.

PR #996, opened 2023-03-04
Support the multiple inputs of keras model.fit.

### LINKED ISSUE #939: How to train model that takes mutiple tensors as input
I code a model for testing which takes 2 differnt tensors as input as follows:
``` 
            Tensor input_wide = keras.Input(5);
            Tensor input_deep = keras.Input(6);

            var hidden1 = keras.layers.Dense(30, activation: keras.activations.Relu).Apply(input_deep);
            var hidden2 = keras.layers.Dense(30, activation: keras.activations.Relu).Apply(hidden1);
            var concat = keras.layers.Concatenate().Apply(new Tensors(input_wide, hidden2));

            var output1 = keras.layers.Dense(1, activation: keras.activations.Relu).Apply(concat);
            var output2 = keras.layers.Dense(1, activation: keras.activations.Relu).Apply(hidden2);

            var model = keras.Model(new Tensors(input_wide, input_deep),
                new Tensors(output1, output2)
            );

            model.compile(optimizer: keras.optimizers.Adam(),
                loss: keras.losses.MeanSquaredError(),
                new[] { "accuracy" });
```
Here I construct a Tensors with 2 tensor, which is assigned to the inputs of model. This model can be compiled without problems. Then I try to predict by this model:
``` 
                // Just test data
                var x1 = np.array(new float[,] { { 1, 2, 3, 4, 5 }, { 1, 3, 5, 7, 9 } });
                var x2 = np.array(new float[,] { { 1, 2, 3, 4, 5, 6 }, { 1, 3, 5, 7, 9, 11 } });
                var x = new Tensors(x1, x2);
                var pred = model.predict(x);
``` 
Here I got an exception `System.Collections.Generic.KeyNotFoundException:“The given key '3' was not present in the dictionary.”`.Did I make a mistake or is it a bug? Many thanks.

AsakusaRinne:
Partially fix #939 

It mainly did two things:
1. Support the multiple inputs of keras `model.fit`.
2. Update the `anonymous_iterator_v2` to `anonymous_iterator_v3`.

It made the following break changes:
1. Removed the implicit conversionfrom `Array` to `NDArray`. Instead, it added a series implicit conversion such as `int[]`, `float[,]`, `double[,,]` and so on.
2. It makes `IDatasetV2` implements `IEnumerable<(Tensors, Tensors)>` instead of `IEnumerable<(Tensor, Tensor)>`. (This may lead to some error of examples.)

### DIFF
--- a/src/TensorFlowNET.Core/Data/IDatasetV2.cs
+++ b/src/TensorFlowNET.Core/Data/IDatasetV2.cs
@@ -4,7 +4,7 @@ using Tensorflow.Framework.Models;
 
 namespace Tensorflow
 {
-    public interface IDatasetV2 : IEnumerable<(Tensor, Tensor)>
+    public interface IDatasetV2 : IEnumerable<(Tensors, Tensors)>
     {
         string[] class_names { get; set; }
 
@@ -17,6 +17,8 @@ namespace Tensorflow
         TensorSpec[] element_spec { get; }
 
         TensorSpec[] structure { get; set; }
+
+        int FirstInputTensorCount { get; set; }
 
         /// <summary>
         /// Caches the elements in this dataset.
--- a/src/TensorFlowNET.Core/NumPy/NDArray.Implicit.cs
+++ b/src/TensorFlowNET.Core/NumPy/NDArray.Implicit.cs
@@ -14,7 +14,76 @@ namespace Tensorflow.NumPy
             red = data[2];
         }
 
-        public static implicit operator NDArray(Array array)
+        public static implicit operator NDArray(int[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(byte[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(float[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(double[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(long[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(bool[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(uint[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(ulong[] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(int[,] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(byte[,] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(float[,] array)
+            => new NDArray(array);
+
+        public static implicit operator NDArray(double[,] array)
+            => new NDArray(array);
+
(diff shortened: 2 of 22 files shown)


src/TensorFlowNET.Core/NumPy/NDArray.Implicit.cs
```diff
@@ -14,7 +14,76 @@ public void Deconstruct(out byte blue, out byte green, out byte red)
             red = data[2];
         }
 
-        public static implicit operator NDArray(Array array)
+        public static implicit operator NDArray(int[] array)
```
Oceania2018, line 17: Is it possible to use like `public static implicit operator NDArray<T>(T[] array)`?

AsakusaRinne: It's attractive but C# does not support using generic implicit operators.🤣

src/TensorFlowNET.Core/Data/IDatasetV2.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Tensorflow
 {
-    public interface IDatasetV2 : IEnumerable<(Tensor, Tensor)>
+    public interface IDatasetV2 : IEnumerable<(Tensors, Tensors)>
```
Oceania2018, line 7: Can we keep both of them in the interface level?
```csharp
    public interface IDatasetV2 : IEnumerable<(Tensor, Tensor)>;
    public interface IDatasetV2 : IEnumerable<(Tensors, Tensors)>;
```

AsakusaRinne: What about keeping it implementing `IEnumerable<(Tensor, Tensor)>` and add an API to get `IEnumerator<(Tensors, Tensors)>`? If the dataset actually holds only `Tensor`, then the tennsor will be casted to `Tensors` when get the `IEnumerator<(Tensors, Tensors)>`.

However, if the dataset actually has multiple inputs as `Tensors`, then it may lead to some confusion when getting the `IEnumerable<(Tensor, Tensor)>`.

src/TensorFlowNET.Keras/Engine/DataAdapters/TensorLikeDataAdapter.cs
```diff
@@ -20,7 +21,7 @@ public TensorLikeDataAdapter(DataAdapterArgs args)
         {
             this.args = args;
             _process_tensorlike();
-            num_samples = (int)args.X.shape[0];
+            num_samples = (int)args.X[0].shape[0];
```
Oceania2018, line 24: Can we use `(int)args.X.shape[0]` instead of `(int)args.X[0].shape[0]`?

AsakusaRinne: Sure, I have changed it.

PR #999, opened 2023-03-04
Reduce the time of keras unittest.

AsakusaRinne:
Now the unit test of keras takes too long, which is caused by the `SavedModel` tests. This PR reduced the data fed to `model.fit` to make the test faster.


test/TensorFlowNET.Keras.UnitTest/SaveModel/SequentialModelSave.cs
```diff
@@ -18,15 +18,15 @@ public class SequentialModelSave
     [TestMethod]
     public void SimpleModelFromAutoCompile()
     {
-        var inputs = tf.keras.layers.Input((28, 28, 1));
-        var x = tf.keras.layers.Flatten().Apply(inputs);
-        x = tf.keras.layers.Dense(100, activation: tf.nn.relu).Apply(x);
-        x = tf.keras.layers.Dense(units: 10).Apply(x);
-        var outputs = tf.keras.layers.Softmax(axis: 1).Apply(x);
-        var model = tf.keras.Model(inputs, outputs);
+        var inputs = keras.layers.Input((28, 28, 1));
```
Oceania2018, line 21: The official example use `tf.keras.xxx` more.

AsakusaRinne: OK, I'll revise it.

PR #1004, opened 2023-03-08
Fix model.evaluate don't have output

### LINKED ISSUE #986: Keras model.evaluate does not have output.
The `model.evaluate` in keras should output the metric of the model to terminal but not.

Wanglongzhi2001:
Fixed #986

### DIFF
--- a/src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
+++ b/src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
@@ -72,12 +72,13 @@ namespace Tensorflow.Keras.Engine
 
                 foreach (var step in data_handler.steps())
                 {
-                    callbacks.on_train_batch_begin(step);
+                    callbacks.on_test_batch_begin(step);
                     logs = test_function(data_handler, iterator);
                     var end_step = step + data_handler.StepIncrement;
                     callbacks.on_test_batch_end(end_step, logs);
                 }
             }
+            Console.WriteLine();
             GC.Collect();
             GC.WaitForPendingFinalizers();
         }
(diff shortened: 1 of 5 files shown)


src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
```diff
@@ -46,18 +55,31 @@ public partial class Model
...
+                //callbacks.on_epoch_begin(epoch);
                 // data_handler.catch_stop_iteration();
-                IEnumerable<(string, Tensor)> results = null;
+                IEnumerable<(string, Tensor)> logs = null;
+
                 foreach (var step in data_handler.steps())
                 {
-                    // callbacks.on_train_batch_begin(step)
-                    results = test_function(data_handler, iterator);
+                    callbacks.on_train_batch_begin(step);
```
AsakusaRinne, line 75: Should this be `on_test_batch_begin`?

Wanglongzhi2001: > Should this be `on_test_batch_begin`?

sry, there are some mistakes happened, I will correct it.

Wanglongzhi2001:
I'm sorry, but I would like to ask about the reason.

src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
```diff
@@ -46,18 +55,32 @@ public partial class Model
...
+                    var end_step = step + data_handler.StepIncrement;
+                    callbacks.on_test_batch_end(end_step, logs);
                 }
             }
+            Console.WriteLine();
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
         }
 
         public KeyValuePair<string, float>[] evaluate(IDatasetV2 x)
```
AsakusaRinne, line 86: This method is an overload of the method above, which should have a similar implementation. I think maybe adding an `EvaluateInternal` is better, just like `FitInternal`.

AsakusaRinne (approved):

Oceania2018 (approved):

PR #1011, opened 2023-03-26
Added: complex, real, imag, angle

BalashovK:
I have added basic missing functionality for complex numbers. Unfortunately, only complex64 is fully functional, complex128 needs more work which I do not know how to do. Please merge or let me now what needs to be adjusted.


src/TensorFlowNET.Core/APIs/tf.complex.cs
```diff
@@ -0,0 +1,26 @@
+﻿/*****************************************************************************
+   Copyright 2018 The TensorFlow.NET Authors. All Rights Reserved.
```
Oceania2018, line 2: Is it better to change to `2023 {Your Name}`

BalashovK: I have moved "complex" to tf.math.cs. I think, it will be more in line with the other code structure. The reason why I created it as a separate file was - I uses "exp" as an example. I would suggest moving "exp" inside of tf.math.cs too.

BalashovK: Please let me know if anything else needs to be changed before the pull request will be pulled. I have a plan to work on adding partial tf.signal fft support, but I need complex numbers support to be in the repository first.

PR #1014, opened 2023-03-26
Fix validation_split has no output and add validation_data parameter to model.fit

Wanglongzhi2001:
Fix validation_split has no output and add validation_data parameter to model.fit and the definition of logs has been modified to align with the log form of train (DIctionary<string, float>)

### DIFF
--- a/src/TensorFlowNET.Core/Keras/Engine/ICallback.cs
+++ b/src/TensorFlowNET.Core/Keras/Engine/ICallback.cs
@@ -15,5 +15,5 @@ public interface ICallback
     void on_predict_end();
     void on_test_begin();
     void on_test_batch_begin(long step);
-    void on_test_batch_end(long end_step, IEnumerable<(string, Tensor)> logs);
+    void on_test_batch_end(long end_step, Dictionary<string, float> logs);
 }
--- a/src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
+++ b/src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
@@ -26,6 +26,7 @@ namespace Tensorflow.Keras.Engine
         /// <param name="workers"></param>
         /// <param name="use_multiprocessing"></param>
         /// <param name="return_dict"></param>
+        /// <param name="is_val"></param>
         public Dictionary<string, float> evaluate(NDArray x, NDArray y,
             int batch_size = -1,
             int verbose = 1,
@@ -33,7 +34,9 @@ namespace Tensorflow.Keras.Engine
             int max_queue_size = 10,
             int workers = 1,
             bool use_multiprocessing = false,
-            bool return_dict = false)
+            bool return_dict = false,
+            bool is_val = false
+            )
         {
             if (x.dims[0] != y.dims[0])
             {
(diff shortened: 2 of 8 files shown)


src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
```diff
@@ -63,31 +66,76 @@ public partial class Model
             });
             callbacks.on_test_begin();
 
-            IEnumerable<(string, Tensor)> logs = null;
+            //Dictionary<string, float>? logs = null;
+            var logs = new Dictionary<string, float>();
```
Wanglongzhi2001, line 70: Because I wasn't very familiar with C# at the time of my last commit, so I change the definition of logs this time to align with the form of Model.fit,

src/TensorFlowNET.Keras/Engine/Model.Fit.cs
```diff
@@ -40,11 +42,17 @@ public partial class Model
...
-            int train_count = Convert.ToInt32(x.dims[0] * (1 - validation_split));
-            var train_x = x[new Slice(0, train_count)];
-            var train_y = y[new Slice(0, train_count)];
-            var val_x = x[new Slice(train_count)];
-            var val_y = y[new Slice(train_count)];
+
+            var train_x = x;
+            var train_y = y;
+
+            if (validation_split != 0f && validation_data == null)
```
Wanglongzhi2001, line 49: When validation_split is not equals to 0 and validation_data is not null, we ignore the validation_split.

src/TensorFlowNET.Keras/Engine/Model.Fit.cs
```diff
@@ -22,13 +22,15 @@ public partial class Model
...
         /// <param name="validation_split"></param>
+        /// <param name="validation_data"></param>
         /// <param name="shuffle"></param>
         public ICallback fit(NDArray x, NDArray y,
             int batch_size = -1,
             int epochs = 1,
             int verbose = 1,
             List<ICallback> callbacks = null,
             float validation_split = 0f,
+            (NDArray val_x, NDArray val_y)? validation_data = null,
```
Wanglongzhi2001, line 33: Add a validation_data parameter to align with python

src/TensorFlowNET.Keras/Engine/Model.Fit.cs
```diff
@@ -110,29 +126,30 @@ public partial class Model
...
         }
 
         public History fit(IDatasetV2 dataset, 
-            IDatasetV2 validation_data = null,
             int batch_size = -1,
             int epochs = 1,
             int verbose = 1,
             List<ICallback> callbacks = null,
-            float validation_split = 0f,
+            //float validation_split = 0f,
```
Wanglongzhi2001, line 144: I haven't figured out how to split when input is IDatasetV2 type

src/TensorFlowNET.Keras/Engine/Model.Fit.cs
```diff
@@ -146,7 +163,10 @@ public partial class Model
                 Model = this,
                 StepsPerExecution = _steps_per_execution
             });
+            foreach( var (x,y) in dataset)
```
Wanglongzhi2001, line 166: oops, this is a typo, I will commit again to delete it.

src/TensorFlowNET.Keras/Engine/Model.Fit.cs
```diff
@@ -193,6 +215,123 @@ public partial class Model
                     {
                         logs["val_" + log.Key] = log.Value;
                     }
+                    callbacks.on_train_batch_end(End_step, logs);
```
Wanglongzhi2001, line 218: Because some new log has been added to the logs after the evaluation, so we need to call it again

src/TensorFlowNET.Core/Keras/Engine/ICallback.cs
```diff
@@ -15,5 +15,5 @@ public interface ICallback
     void on_predict_end();
     void on_test_begin();
     void on_test_batch_begin(long step);
-    void on_test_batch_end(long end_step, IEnumerable<(string, Tensor)> logs);
+    void on_test_batch_end(long end_step, Dictionary<string, float> logs);
```
Wanglongzhi2001, line 18: Since I rectify the definition of logs in Model.Evaluate, so here and somewhere need to change either.

src/TensorFlowNET.Keras/Engine/Model.Evaluate.cs
```diff
@@ -26,14 +26,17 @@ public partial class Model
...
         public Dictionary<string, float> evaluate(NDArray x, NDArray y,
             int batch_size = -1,
             int verbose = 1,
             int steps = -1,
             int max_queue_size = 10,
             int workers = 1,
             bool use_multiprocessing = false,
-            bool return_dict = false)
+            bool return_dict = false,
+            bool is_val = false
```
Wanglongzhi2001, line 38: Because when we call evaluate using validation_split in model.fit, we don't need to call the on_test_batch_end, which will interfere with our output on the screen. And when we call model.evaluate , we does need the on_test_batch_end. So I pass a parameter to distinguish between the two cases

src/TensorFlowNET.Keras/Engine/Model.Fit.cs
```diff
@@ -193,6 +215,123 @@ public partial class Model
...
+                callbacks.on_epoch_end(epoch, logs);
+
+                GC.Collect();
+                GC.WaitForPendingFinalizers();
+            }
+
+            return callbacks.History;
+        }
+
+        History FitInternal(DataHandler data_handler, int epochs, int verbose, List<ICallback> callbackList, (IEnumerable<Tensor>, NDArray)? validation_data,
```
Wanglongzhi2001, line 290: I overload FitInternal to handle different types of validation_data

Wanglongzhi2001:
And this is the result:

1%LSUFG`TJ37CW}WS8KCS3G

Wanglongzhi2001:
BTW, I often come across model.fit does not converge, this has been bothering me for a long time, but I have not found the reason.

Oceania2018 (approved):

Issue #1012, opened 2023-03-26
[BUG Report]: Conv2DTranspose not working as expected


PhasonMatrix:
### Description

I am implementing a U-Net convolutional network. I have some working Python code which I am now translating over to C#. When I try to use Tensorflow.Keras.ILayer.Conv2DTranspose() I have encountered two problems.

Problem 1: If I don't include the activation parameter, or set the activation parameter to null, I get an ArgumentNullException, with the exception message as "Value cannot be null. (Parameter 'key')". The activation parameter has a default value of null so it should be okay with this value. In my python code I do not get any error when not including a value for this parameter. Also, the error message says that the missing parameter is called 'key', which is incorrect, or maybe it's referring to some other parameter further down the call stack.

Example code:
`var u6Transpose = keras.layers.Conv2DTranspose(128, kernel_size: (2, 2), strides: (2, 2), output_padding: "same");`

Problem 2: After successfully constructing a Conv2DTranspose object, when calling the Apply() method, I get a NotImplementedException. 

Example code:
`var u6Tensor = u6Transpose.Apply(c5PostDropout);`

Please let me know if this is not the correct way to use this layer. I have not found any example code online.

### Configuration and Other Information

.Net 6.0
Compile target: x64

NuGet packages:
TensorFlow.NET 0.100.4
TensorFlow.Keras 0.10.4
SciSharp.TensorFlow.Redist 2.11.0

PhasonMatrix:
Regarding Problem 1. I think I know why the error message says the parameter name is 'key'. In Activations.cs: line 80, `Activations.GetActivationFromName()` passes the name to `_nameActivationMap.TryGetValue()`. _nameActivationMap is a dictionary, and so doesn't like you trying to get a value with a null key. Maybe `Activations.GetActivationFromName()` should return null if the `name` parameter is null.

PhasonMatrix:
I have tested again with nightly build to see if the problem has been fixed. Unfortunately it has not. Problem 2 is still behaving the same. I can see that there is a fix for calling len() on KerasShapesWrapper (#1078), but I am still getting the exception "Message=len() not implemented for type: Tensorflow.Keras.Saving.KerasShapesWrapper". Should the nightly build "0.100.5-BERT-load-33-g89f3b238-nightly" have the fix?

lingbai-kong:
No, that nightly build hasn't included the fix. I think the next version will release the bug-fixing commit.

[...]

PR #1076, opened 2023-05-17
fix: error when set the activation of keras.layers.Conv2DTranspose to null.

lingbai-kong:
Solve 1st problem of #1012

The bug is caused by the unchecked variable in GetActivationFromName().

In original TensorFlow, the corresponding method just [returns the 'linear' function](https://github.com/tensorflow/tensorflow/blob/f5157cf2fefe18ffd82a50df4b1d18fd39631123/tensorflow/python/keras/activations.py#LL594C5-L594C12) when encountering this null situation. 

I solved the problem in their way.

Thanks!


Oceania2018:
Thanks for your contribution, keep it up!

PR #1079, opened 2023-05-19
fix: status null reference of base session.

AsakusaRinne:


AsakusaRinne:
partially fix the issue #1077

PR #1085, opened 2023-05-24
feat: add Activation Adaptor

lingbai-kong:
I have created a new class called ```ActivationAdaptor```, which enables Layers to accept different types of **activation** parameters, such as the **string** "tanh", **Activation**, and **Func**. This also provides convenience for adding more types of activations in the future, which avoids redunctunt overloaded function and ensures clarity of the source code.
I'm not sure if this PR is suitable, I would appreciate it if you could carefully review the differences and provide any necessary feedback.
By the way, if this PR is approved, I plan to add similar adaptors for **Initializer** and **IRegularizer**.
Thanks!


src/TensorFlowNET.Core/Keras/Layers/ILayersApi.cs
```diff
@@ -48,7 +48,7 @@ public partial interface ILayersApi
             string data_format = "channels_last",
             int dilation_rate = 1,
             int groups = 1,
-            string activation = null,
```
AsakusaRinne, line 51: I think we should keep the function with `string` as the type of activation, because using string to specify an activation is quite common in tensorflow. Though the implicit conversion from `string` to `ActivationAdapter` has been provided, it may cause some confusion to users. Maybe splitting them to two functions is better, the one with `string` and the other with `ActivationAdapter`.

src/TensorFlowNET.Core/Keras/Activations/Activations.cs
```diff
@@ -25,10 +26,52 @@ public class Activation
             };
         }
     }
+    public class ActivationAdaptor
```
AsakusaRinne, line 29: Providing some constructors and adding descriptions for usage is better, because one of the properties must be specified when using it.

lingbai-kong: Thank you for your feedback. I will make the suggested improvements and resubmit the pull request for your review.

AsakusaRinne: After opening a pull request, the commits after that will update the status of PR. Therefore when in review, just making an extra commit is ok. :)

lingbai-kong: Got it!✔

src/TensorFlowNET.Core/Keras/Activations/Activations.cs
```diff
@@ -26,9 +27,65 @@ public class Activation
         }
     }
 
+    /// <summary>
+    /// The ActivationAdaptor is used to store string, Activation, and Func for Laysers Api to accept different types of activation parameters.
+    /// One of the properties must be specified while initializing.
+    /// </summary>
+    public class ActivationAdaptor
```
Oceania2018, line 34: It should be `ActivationAdapter`.

src/TensorFlowNET.Core/Keras/Activations/Activations.cs
```diff
@@ -26,9 +27,65 @@ public class Activation
...
+        /// The Func definition of activation function, which can be customized.
+        /// </summary>
+        public Func<Tensor, string, Tensor>? Func { get; set; } = null;
+
+        public ActivationAdaptor(string name)
+        {
+            Name = name;
+        }
+
+        public ActivationAdaptor(Activation activation)
```
Oceania2018, line 56: Check if we can use `IActivation`.

src/TensorFlowNET.Core/Keras/Activations/Activations.cs
```diff
@@ -26,9 +27,65 @@ public class Activation
...
+    {
+        /// <summary>
+        /// The name of activaiton function, such as `tanh`, `sigmoid`.
+        /// </summary>
+        public string? Name { get; set; } = null;
+
+        /// <summary>
+        /// The available Activation instance of activaiton function, such as keras.activations.Tanh, keras.activations.Sigmoid.
+        /// </summary>
+        public Activation? Activation { get; set; } = null;
```
Oceania2018, line 44: `IActivation` is the interface of any `Activation`.

lingbai-kong: Thanks for figuring out these mistakes. I'll resubmit soon. : )

lingbai-kong:
There is a failure in linux build and test. The log does not shows any useful information but a error code `MSB4166`. The [stackoverflow](https://stackoverflow.com/questions/55070502/intermittent-build-break-with-msb4166-child-node-exited-prematurely-vs-15/55209164#55209164) said the cause is not sure but add the /nodeReuse:false to the msbuild command can help fix this issue. However, the build and test passed on Windows. So... does it matter?

Oceania2018 (changes requested):

PR #1089, opened 2023-06-01
refactor: Standardize TensorFlowNET.Keras/Losses/*

DevNullx64:
abstract Loos.Apply instead of virtualy throw exception
abstract class LossFunctionWrapper
Smooth implementation of sub class
Move ILossFunc to base abstract class Loss


Oceania2018:
Why do we remove `ILossFunc`? It's designed to decouple the `Tensorflow.Binding` with `Tensorflow.Keras` projects.
`ILossFunc` is the standard interface between the two projects.
I think what we need to do is changing all the `loss` function to implement the interface, then in `Tensorflow.Binding` can use `Keras` loss functions.

DevNullx64:
I've put the ILossFunc implementation on base Loss abstract class. Everithing that inherit from Loss inherit ILossFunc.

Issue #635, opened 2020-11-05
NDArray.max not accepting int array of axis


Banyc:
`NDArray.max()` should support max from multiple dimensions.

However, the current version `0.21.0` of Tensorflow.NET only supports one dimension.

Banyc:
In addition, the `NDArray.sum()` method got the same issue.

PR #1091, opened 2023-06-02
fix: A new feature for NDArray.max 

RayWang-iat:
Solve the problem that NDArray.max() does not support max from multiple dimensions and add a TestMethod. #635 

And the NDArray.sum() method has already supported max from multiple dimensions.

Thanks!

### DIFF
--- a/test/TensorFlowNET.UnitTest/Numpy/Math.Test.cs
+++ b/test/TensorFlowNET.UnitTest/Numpy/Math.Test.cs
@@ -65,5 +65,17 @@ namespace TensorFlowNET.UnitTest.NumPy
             var y = np.power(x, 3);
             Assert.AreEqual(y, new[] { 0, 1, 8, 27, 64, 125 });
         }
+         [TestMethod]
+        public void maximum()
+        {
+            var x1 = new NDArray(new[,] { { 1, 2, 3 }, { 4, 5.1, 6 } });
+            var x2 = new NDArray(new[,] { { 3, 2, 1 }, { 6, 5.1, 4 } });
+            var y = np.maximum(x1,x2);
+            var y1 = np.maximum(x1, x2, axis: 0);
+            var y2 = new NDArray(new[,] { { 3, 2, 3 }, { 6, 5.1, 6 } });
+            var y3 = new NDArray(new[] {  6, 5.1, 6 });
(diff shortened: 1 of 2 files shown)


test/TensorFlowNET.UnitTest/Numpy/Math.Test.cs
```diff
@@ -65,5 +65,17 @@ public void power()
             var y = np.power(x, 3);
             Assert.AreEqual(y, new[] { 0, 1, 8, 27, 64, 125 });
         }
+         [TestMethod]
+        public void maximum()
+        {
+            var x1 = new NDArray(new[,] { { 1, 2, 3 }, { 4, 5.1, 6 } });
+            var x2 = new NDArray(new[,] { { 3, 2, 1 }, { 6, 5.1, 4 } });
+            var y = np.maximum(x1,x2);
+            var y1 = np.maximum(x1, x2, axis: 0);
```
Oceania2018, line 74: Thanks for your contribution.
Can you add one more testing case
```csharp
np.maximum(x1, x2, axis: 1)
```

test/TensorFlowNET.UnitTest/Numpy/Math.Test.cs
```diff
@@ -66,16 +66,19 @@ public void power()
...
             var y1 = np.maximum(x1, x2, axis: 0);
-            var y2 = new NDArray(new[,] { { 3, 2, 3 }, { 6, 5.1, 6 } });
-            var y3 = new NDArray(new[] {  6, 5.1, 6 });
-            Assert.AreEqual(y, y2);
-            Assert.AreEqual(y1, y3);
+            var y2 = np.maximum(x1, x2, axis: 1);
+            var y3 = new NDArray(new[,] { { 3, 2, 3 }, { 6, 5.1, 6 } });
+            var y4 = new NDArray(new[] {  6, 5.1, 6 });
+            var y5 = new NDArray(new[] { 3.0, 6 });
+            Assert.AreEqual(y0, y3);
```
Oceania2018, line 79: We can use the simplier syntax like: `Assert.AreEqual(new[,] { { 3, 2, 3 }, { 6, 5.1, 6 } }, y0.numpy());`

Oceania2018 (approved):

PR #1092, opened 2023-06-04
refactor: Model.eval

DevNullx64:


Wanglongzhi2001:
LGTM，Model.Evaluate does have some repetitive code. For example, test_function and test_step_multi_inputs_function are actually doing the same thing, this modification can reduce the repetitive code of overloading.

Wanglongzhi2001:
BTW, in the test_step_multi_inputs_function of Model.Evaluate, should train_step be changed to test_step,? can you help change it?

Wanglongzhi2001:
Looks good to me.

Wanglongzhi2001:
@DevNullx64 , Seems there exists some bug in the model.evaluate, If the input data is NDArray type, it will matchs both evaluate(Tensor,..) and evaluate(IEnumable<Tensor>,... signatures, which will result in an ambiguous error in the call model.evaluate. The following is a example:
```
            var x_train = np.array(new float[,] { { 0.1f }, { 0.2f }, { 0.3f }, { 0.4f } });
            var y_train = np.array(0, 1, 2, 3);

            var x_test = x_train;
            var y_test = y_train;

            var model = keras.Sequential(new List<ILayer>() {
                tf.keras.layers.Dense(128, activation: "relu", input_shape: new Shape(1)),
                tf.keras.layers.Dense(4)
            });

            var loss_fn = tf.keras.losses.SparseCategoricalCrossentropy(from_logits: true);

            model.compile(optimizer: new Tensorflow.Keras.Optimizers.Adam(),
                loss: loss_fn,
                metrics: new[] { "accuracy" });

            model.fit(x_train, y_train, epochs: 10, verbose: 0);

            model.evaluate(x_test, y_test, verbose: 2);
```

Issue #1088, opened 2023-05-30
[Question]: The Tensors class should not implement an implicit cast to Tensor


DevNullx64:
### Description

The Tensors class implements an implicit cast to Tensor. It simply returns the first Tensor in its list. There is a potential risk of information loss in this case. While casting from float to double does not pose any problems, casting from double to float requires an explicit conversion.

### Alternatives

I looked into the impact of an explicit cast. I made the necessary corrections to pass the tests. However, I had to modify the unit tests to ensure proper casting of Tensors. It doesn't feel clean to do that...

I went a bit further and realized that almost all implementations of the layers use a Tensor as input and output. I got overwhelmed by the task of revising everything quickly. Then, I reached the RNN section, which seems to be under development. At that point, I couldn't make the necessary corrections or add a comment to guide the changes. I gave up on completing the task when I realized the extent of the impact. I was in the process of making a modification that would affect everything derived from Loss or Layer, and with generics (to avoid having to declare typed arguments and hiding inherited members, etc.).

AsakusaRinne:
@Oceania2018 any ideas about this?

DevNullx64:
There is nothing insurmountable, I believe. Classes working with a tensor and returning a tensor should have a common abstract class inheriting from Layer. It would virtually and naively implement the Call method for each tensor input and associate it with its output.
There are few layers that really need to implement tensors as inputs. They could keep their own implementations.

My main concern is that the modification will still change a lot of things in the keras.layer directory of the TensorFlow.Keras project.

I'm not an expert in TensorFlow or Keras, so forgive me if I'm mistaken.

AsakusaRinne:
Thank you a lot for telling us this concern. I agree that the implicit cast from Tensors to Tensor is risky. To me this proposal hits a key problem and sounds very reasonable. I approve removing this implicit conversion in the future. Here're some of my thoughts of the design of `Tensors` and welcome to discuss it together.

1. The `Apply` of the layer, which is corresponding to `__call__` in python, should keep using `Tensors` because it will make APIs consistent. Otherwise users may need to figure out whether the `Apply` with `Tensors` or the one with `Tensor` should be used for some layers.
2. Originally I wanted to delay this change to v1.0.0 to avoid break change in minor version. However when developing `RNN` recently, I found it better to change some behaviors of `Tensors`. Therefore I wonder if making this modification in `v0.100.6` will impact much on users.
[...]

PR #1097, opened 2023-06-07
feat: add rnn basic modules

AsakusaRinne:
1. Refactor `Tensors` to make it support nested structure.
2. Add `Nest<T>` and `INestStructure<T>` to enable better use of nest structures.
3. Add basic RNN modules.
4. Revise the implementation of `TensorArray`.

Note:
The `ILayer.Call` needs to be fully refactored in the future.

@DevNullx64 This PR partially refactors the `Tensors`, which is mentioned in #1088 . Please feel free to comment anything about that :). Since the target branch is a dev branch, this PR will be merged at once. But if there's disadvantages you catch, we could improve it before merging to master branch.

In this PR the implicit conversion from `Tensors` to `Tensor` is still kept because the main purpose of this PR is to support RNN. However I add a check when doing this implicit conversion. If a `Tensors` has more than one `Tensor`, then an exception will be thrown. In this way, some error will be visiable.


DevNullx64:
Quite not sure, but I think that by adding some class or interface, Nest<T> can hold less. By exemple what is done with the NestType may, prehaps, done by splitting or generic type handling. I seen that "NestType.Node" can't use PackSequenceInternal(). Change class architecture can prehaps avoid possibility to call PackSequenceInternal() from a Nest<T> that should currently have NestType.Node.

I hope to don't borring you to many, I will help you if you want.

After that, you've done a lot of work :) !
I'm loking for this pull.

PR #1105, opened 2023-06-14
Fix bug: fix the bug of generate_zero_filled_state when in graph mode

Wanglongzhi2001:


Oceania2018:
Thanks for your contribution.

PR #1110, opened 2023-06-18
feat: Support training of RNN and LSTM.

Wanglongzhi2001:
1. Support the model building of RNN.
2. Support the model building of LSTM.
3. Support the training of RNN.
4. Support the training of LSTM.
5. Add test.
6. Refactor Tensors class to deal with implicit conversion from Tensors to Tensor and support Nest Tensors.
7. Add Nest class to handle nested Tensor cases, and add corresponding Flatten and Mapstructure API.
8. Use IOptionalArgs in Layer.Apply to prevent the inconsistency of Call method parameters in different layers, resulting in the need to set too many Call method parameters.
9. Add the implemention of while_loop in graph mode.


Wanglongzhi2001:
It builds on windows successfully, but fails to build on linux. According to the error message, it seems that the problem is on protobuf. Is this matters? @AsakusaRinne

AsakusaRinne:
> It builds on windows successfully, but fails to build on linux. According to the error message, it seems that the problem is on protobuf. Is this matters? @AsakusaRinne

I'll test it on linux later

AsakusaRinne:
The problem in ci has been located, which will be solved after upgrading the redist packages later.

PR #1111, opened 2023-06-19
feat: np update square and dot product

visagang:


Oceania2018 (approved):
Thanks for your contribution.

PR #1114, opened 2023-06-22
Improve the API of  LayerNormalization

Beacontownfc:
The ` LayerNormalization` didn't work as expected, it throws an `System.NullReferenceException` during runtime.

### DIFF
--- a/src/TensorFlowNET.Core/APIs/tf.nn.cs
+++ b/src/TensorFlowNET.Core/APIs/tf.nn.cs
@@ -17,7 +17,6 @@
 using System.Xml.Linq;
 using Tensorflow.Operations;
 using Tensorflow.Operations.Activation;
-//using static System.Formats.Asn1.AsnWriter;
 using static Tensorflow.Binding;
 
 namespace Tensorflow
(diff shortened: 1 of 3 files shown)


src/TensorFlowNET.Core/APIs/tf.nn.cs
```diff
@@ -14,8 +14,10 @@
    limitations under the License.
 ******************************************************************************/
 
+using System.Xml.Linq;
 using Tensorflow.Operations;
 using Tensorflow.Operations.Activation;
+//using static System.Formats.Asn1.AsnWriter;
```
AsakusaRinne, line 20: Is this using necessary?

src/TensorFlowNET.Core/APIs/tf.nn.cs
```diff
@@ -125,6 +127,22 @@ public Tensor relu(Tensor features, string name = null)
                     is_training: is_training,
                     name: name,
                     exponential_avg_factor: exponential_avg_factor);
+            public Tensor batch_normalization(Tensor x,
```
AsakusaRinne, line 130: Please add some simple comments of this API since it's exposed to users

AsakusaRinne (changes requested):

src/TensorFlowNET.Core/APIs/tf.nn.cs
Beacontownfc: Not necessary, deleted.

src/TensorFlowNET.Core/APIs/tf.nn.cs
Beacontownfc: Done

AsakusaRinne (approved):

PR #1129, opened 2023-07-03
fix: optimize some APIs

Beacontownfc:
`tf.nn.moments` : There may be a problem where the gradient is `null` during the training,  it seems that removing the 681st line of code from the `[TensorFlowNET.Core/Operations/array_ops.cs]` can solve this issue.

`tf.nn.batch_normalization`: Use `nn_impl.batch_normalization` to reuse code.

### DIFF
--- a/src/TensorFlowNET.Core/Operations/array_ops.cs
+++ b/src/TensorFlowNET.Core/Operations/array_ops.cs
@@ -678,7 +678,6 @@ namespace Tensorflow
             var tape = tf.GradientTape().stop_recording();
             var result = gen_array_ops.stop_gradient(input, name);
             tape.StartRecord();
-            tf.GradientTape().PushTape(tape);
             return result;
         }
 
--- a/src/TensorFlowNET.Core/APIs/tf.nn.cs
+++ b/src/TensorFlowNET.Core/APIs/tf.nn.cs
@@ -144,16 +144,8 @@ namespace Tensorflow
                         Tensor offset,
                         Tensor scale,
                         float variance_epsilon,
-                        string name = null)
-            {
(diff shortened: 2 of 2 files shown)


src/TensorFlowNET.Core/Operations/array_ops.cs
```diff
@@ -678,7 +678,6 @@ public static Tensor stop_gradient(Tensor input, string name = null)
             var tape = tf.GradientTape().stop_recording();
             var result = gen_array_ops.stop_gradient(input, name);
             tape.StartRecord();
-            tf.GradientTape().PushTape(tape);
```
OliBomby, line 681: this removal looks highly suspicious. Does anyone even know why this statement is here?

Oceania2018: Everything looks fine when I run though all the [examples](https://github.com/SciSharp/SciSharp-Stack-Examples).

OliBomby: none of the examples test this specific method though

Beacontownfc: Not deleting this line of code will result in `NULL` gradient when using the `stop_gradient` API.

Beacontownfc: [LayerNorm](https://github.com/SciSharp/TensorFlow.NET/blob/master/src/TensorFlowNET.Keras/Layers/Normalization/LayerNormalization.cs) uses `tf.nn.moment` and `tf.nn.moment` uses the `stop_gradient` API. If this line of code is not deleted, an error will be reported. Deleting this line of code will enable normal training.

PR #1160, opened 2023-07-26
add image save as jpg and png funcion and testunit

dogvane:


Oceania2018:
Please fix the UnitTest
```shell
Test run for D:\a\TensorFlow.NET\TensorFlow.NET\test\TensorflowNET.Hub.Unittest\bin\Debug\net6\Tensorflow.Hub.Unittest.dll (.NETCoreApp,Version=v6.0)
  Failed ImageSaveTest [14 s]
  Error Message:
   Test method TensorFlowNET.UnitTest.ImageTest.ImageSaveTest threw exception: 
Tensorflow.ValueError: Cannot use the default session to execute operation: the operation's graph is different from the session's graph. Pass an explicit session to run(session=sess).
  Stack Trace:
      at Tensorflow.ops._run_using_default_session(Operation operation, FeedItem[] feed_dict, Graph graph, Session session) in D:\a\TensorFlow.NET\TensorFlow.NET\src\TensorFlowNET.Core\ops.cs:line 505
```

PR #1164, opened 2023-08-26
fix: partially fix the bug of load_model

Wanglongzhi2001:
Load_model use reflection to revive LayerArgs and then revive Layer.
https://github.com/SciSharp/TensorFlow.NET/blob/adeed05f64cf033a7d945d97ba4b59bc5ec4aa03/src/TensorFlowNET.Keras/Utils/generic_utils.cs#L64-L69
So each layer's args must be the pattern of {class_name}Args, otherwise it cannot be revived.


Wanglongzhi2001:
But there still exists some bug when usr Merge Layer such as Add layer or Substract layer. When use Merge Layer, there will be some error when buildMap. So user arr better to not use Merge Layer at present.

Issue #1170, opened 2023-09-10
[BUG Report]: EarlyStop does not stop


asaf92:
### Description

When using EarlyStop and reaching a point where the training should stop, it just repeats the message `Restoring model weights from the end of the best epoch:` but keeps training. 

It looks like it sets `stop_training = true` correctly, but the value of stop_training is not checked anywhere in `FitInternal`

Oceania2018:
Released v0.110.3. Can you check if this version works?

Wanglongzhi2001:
Can this issue be closed? @asaf92

GaijinOtohp:
> Released v0.110.3. Can you check if this version works?

The issue still exists even in v0.110.3. The model keeps training while "EarlyStopping" does trigers it by changing "Stop_training" to true.

asaf92:
It still exists in v0.110.3 @Wanglongzhi2001 

I decompiled and saw that the code wasn't changed

PR #1171, opened 2023-09-10
fix: model does not stop on stop_training == true #1170

asaf92:
Did not test it, but debugging showed that this is the code that is missing. `stop_training` is only updated but never used.

This solves #1170

### DIFF
--- a/src/TensorFlowNET.Keras/Engine/Model.Fit.cs
+++ b/src/TensorFlowNET.Keras/Engine/Model.Fit.cs
@@ -224,6 +224,10 @@ namespace Tensorflow.Keras.Engine
 
                 GC.Collect();
                 GC.WaitForPendingFinalizers();
+                if (stop_training)
+                {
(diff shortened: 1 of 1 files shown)


Wanglongzhi2001:
LGTM

asaf92:
Hey @Wanglongzhi2001, thank you for accepting my PR!
I noticed that the automatic nightly release has not worked for the last ~2 months. Is there a release planned in the near future where this change will be included?

Oceania2018:
Released v0.110.3. Can you check if this version works?

GaijinOtohp:
> Released v0.110.3. Can you check if this version works?

It seems to work if "Tensorflow.Keras.dll" is directly generated from the current version of the repository. But not from the released v0.110.3.

asaf92:
@Oceania2018 I decompiled with JetBrains and the changes are not included in the release

PR #1172, opened 2023-09-13
cached_session for graph tests

novikov-alexander:
Implementation of cached_session for the possibility of further porting of tests. 

cached_session is a singleton which is constructed on the first call and disposed on the tests cleanup.

There's also testBoundaryContinue() test implemented to prove that cache_session() works.

### DIFF
--- a/test/TensorFlowNET.Graph.UnitTest/ControlFlowTest/WhileContextTestCase.cs
+++ b/test/TensorFlowNET.Graph.UnitTest/ControlFlowTest/WhileContextTestCase.cs
@@ -24,13 +24,13 @@ namespace TensorFlowNET.UnitTest.ControlFlowTest
         private void _testWhileContextHelper(int maximum_iterations)
         {
             // TODO: implement missing code dependencies
-            var sess = this.cached_session();
+            using var sess = this.cached_session();
             var i = constant_op.constant(0, name: "i");
             var c = new Func<Tensor, Tensor>(x => gen_math_ops.less(x, ops.convert_to_tensor(10), name: "c"));
             var b = new Func<Tensor, Tensor>(x => math_ops.add(x, 1, name: "c"));
             //control_flow_ops.while_loop(
             //      c, b, i , maximum_iterations: tf.constant(maximum_iterations));
-            foreach (Operation op in sess.Single().graph.get_operations())
+            foreach (Operation op in sess.graph.get_operations())
             {
                 var control_flow_context = op._get_control_flow_context();
                 /*if (control_flow_context != null)
--- a/test/TensorFlowNET.Graph.UnitTest/ControlFlowTest/WhileContextTestCase.cs
+++ b/test/TensorFlowNET.Graph.UnitTest/ControlFlowTest/WhileContextTestCase.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
+using System.Linq;
 using Tensorflow;
 using static Tensorflow.Binding;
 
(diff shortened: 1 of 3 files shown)


test/TensorFlowNET.Graph.UnitTest/ControlFlowTest/WhileContextTestCase.cs
```diff
@@ -29,7 +30,7 @@ private void _testWhileContextHelper(int maximum_iterations)
             var b = new Func<Tensor, Tensor>(x => math_ops.add(x, 1, name: "c"));
             //control_flow_ops.while_loop(
             //      c, b, i , maximum_iterations: tf.constant(maximum_iterations));
-            foreach (Operation op in sess.graph.get_operations())
+            foreach (Operation op in sess.Single().graph.get_operations())
```
novikov-alexander, line 33: @Oceania2018 I'm not sure how exactly IEnumerable should work there since Python generator produces only one instance. Anyway this test wasn't working and it is still not implemented yet.

test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
```diff
@@ -388,22 +388,19 @@ public void testBoundaryStop()
...
-            //    x = constant(1.0)
-            //    y = x * 2.0
-            //    z = y * 3.0
-            //    grads = gradients.gradients(z, [x, y])
-            //    self.assertTrue(all(x is not None for x in grads))
-            //    self.assertEqual(6.0, grads[0].eval())
+            // Test that we differentiate both 'x' and 'y' correctly when x is a
+            // predecessor of y.
 
+            self.cached_session();
```
novikov-alexander, line 397: @Oceania2018 I'm not sure how exactly "with" should be ported there, since "IEnumerable" isn't "IDisposable" and we don't need to call any "Dispose" for the enumerable itself. I guess session's "Dispose" should be called automatically.

test/TensorFlowNET.Graph.UnitTest/ControlFlowTest/WhileContextTestCase.cs
Oceania2018: The test didn't come through.
```shell
   Test method TensorFlowNET.UnitTest.Gradient.GradientTest.testBoundaryContinue threw exception: 
Tensorflow.ValueError: Cannot use the default session to evaluate tensor: the tensor's graph is different from the session's graph. Pass an explicit session to `eval(session=sess)`.
```

novikov-alexander: Sorry, it was flaky. Now it should work.

Wanglongzhi2001: > @Oceania2018 I'm not sure how exactly IEnumerable should work there since Python generator produces only one instance. Anyway this test wasn't working and it is still not implemented yet.
Since the different behavior of `yield` in Python and C#, If cached_session only need to have one session, I think it is better not to use `yield`? And then you can use `using` to replace `with` in GradientTest.cs

Wanglongzhi2001: > @Oceania2018 I'm not sure how exactly IEnumerable should work there since Python generator produces only one instance. Anyway this test wasn't working and it is still not implemented yet.
Since the different behavior of `yield` in Python and C#, If cached_session only need to have one session, I think it is better not to use `yield`? And then you can use `using` to replace `with` in GradientTest.cs

novikov-alexander: Could, please, clarify why `yield` used in Python in this case?
Anyway fixed.

PR #1173, opened 2023-09-14
Gradient tests which were dependant only on cached_session

novikov-alexander:
testAggregationMethodAccumulateN
testAggregationMethodAddN
testAggregationMethodTree
testSingletonIndexedSlices - **broken**
testDependentYs
testPartialDerivatives
testStopGradients - **broken**

### DIFF
--- a/test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
+++ b/test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
@@ -488,6 +488,7 @@ namespace TensorFlowNET.UnitTest.Gradient
             //    self.assertTrue(isinstance(grads[0], ops.Tensor))
         }
 
+        [Ignore("FIXME")]
         [TestMethod]
         public void testSingletonIndexedSlices()
         {
--- a/test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
+++ b/test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
@@ -488,17 +488,20 @@ namespace TensorFlowNET.UnitTest.Gradient
             //    self.assertTrue(isinstance(grads[0], ops.Tensor))
         }
 
-        [Ignore("FIXME")]
+        [Ignore("TODO: CompositeTensors are not supported yet.")]
         [TestMethod]
         public void testSingletonIndexedSlices()
         {
             tf.Graph().as_default();
 
+            // TODO: uncomment when CompositeTensors are supported.
+            /*
             var x = tf.placeholder(TF_DataType.TF_FLOAT);
             var y = tf.identity(x);
             var dy_indices = tf.placeholder(TF_DataType.TF_INT32);
             var dy_values = tf.placeholder(TF_DataType.TF_FLOAT);
-            Tensor dy = new IndexedSlices(dy_values, dy_indices);
+            var dy = new IndexedSlices(dy_values, dy_indices);
+           
             var dx = tf.gradients(new[] { y }, new[] { x }, grad_ys: new[] { dy })[0];
             // The IndexedSlices gradient of tf.identity is the identity map.
             using (var sess = self.cached_session())
@@ -514,6 +517,7 @@ namespace TensorFlowNET.UnitTest.Gradient
                 var vdy = result[1];
                 self.assertEqual(vdx, vdy);
             }
+            */
 
         }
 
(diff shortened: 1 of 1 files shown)


test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
```diff
@@ -490,24 +488,33 @@ public void testNoGradientForStringOutputs()
...
         [TestMethod]
         public void testSingletonIndexedSlices()
         {
+            tf.Graph().as_default();
+
+            var x = tf.placeholder(TF_DataType.TF_FLOAT);
+            var y = tf.identity(x);
+            var dy_indices = tf.placeholder(TF_DataType.TF_INT32);
+            var dy_values = tf.placeholder(TF_DataType.TF_FLOAT);
+            Tensor dy = new IndexedSlices(dy_values, dy_indices);
```
novikov-alexander, line 501: @Wanglongzhi2001 this one fails because `IndexedSlices` has `null` shape. Should I create an issue?

test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
```diff
@@ -699,15 +782,21 @@ public void testUnconnectedGradientsZerosUnconnectedGradients()
             //        [y], [x], unconnected_gradients="zero")
             //    with self.cached_session() as sess:
             //      self.assertAllEqual([[0.0, 0.0], [0.0, 0.0]], self.evaluate(grads)[0])
+
+            // tf.Graph().as_default();
+            // var x = tf.constant(1.0, shape: new long[] { 2, 2 });
+            // var y = tf.constant(3.0, shape: new long[] { 3, 1 });
+            // var grads = tf.gradients(new[] { y }, new[] { x }, unconnected_gradients: "zero");
```
novikov-alexander, line 789: @Wanglongzhi2001 there're no `unconnected_gradients` parameter here yet.

test/TensorFlowNET.Graph.UnitTest/GradientTest/GradientTest.cs
Wanglongzhi2001: After comparing the implementations of TensorFlow and TensorFlow.NET, it seems that TensorFlow.NET just simpley does not throw error when dense_shape is null. 
https://github.com/tensorflow/tensorflow/blob/704c1918a05574215d0499a74905f9a43fab01d5/tensorflow/python/framework/indexed_slices.py#L435-L438
I'm sorry that I'm not very familiar with tf1, is it correct to use IndexSlice in your example? Or users need to manually pass in the dense_shape parameter? After all, in most case in tensorflow, dense_shape is manually passed in.
https://github.com/tensorflow/tensorflow/blob/704c1918a05574215d0499a74905f9a43fab01d5/tensorflow/python/client/session_test.py#L884-L920

novikov-alexander: I just ported an existing test from TensorFlow. You may that this code was commented before. 
The problem is in conversion from `IndexedSlices` to `Tensor`. Python TensorFlow doesn't have any restrictions on gradients interface and it looks like it converts `IndexedSlices` internally in a different way. I'm trying to run Python TensorFlow tests now to compare, but it is not easy to configure them for the first time.

novikov-alexander: This is what's really going on here:
`IndexedSlices` actually is a `CompositeTensor`.
If we go to `_GradientsHelper` we'll find out that it shoud be able to convert `CompositeTensor` to `Tensor` by its own rules.
https://github.com/tensorflow/tensorflow/blob/8fcc1b465a354c2277dc84b7503443013c693541/tensorflow/python/ops/gradients_util.py#L535
In other words `gradients` conditionally should support the following signature:
```
public (Tensor | CompositeTensor)[] gradients((Tensor | CompositeTensor)[] ys,
    (Tensor | CompositeTensor)[] xs,
    (Tensor | CompositeTensor)[] grad_ys = null,
    ...);
```
Where `(Tensor | CompositeTensor)` is an abstract union type.
However, it is not implemented yet. 
https://github.com/SciSharp/TensorFlow.NET/blob/9ddff6996381b93c5eb285a9d39c6cd67f2122e8/src/TensorFlowNET.Core/Gradients/gradients_util.cs#L48
The code is just skipped.
So, I just commented and marked the test as "TODO".
I can also leave a comment in `gradients` if necessary.

Wanglongzhi2001: I'm agree with you. And it seems that the implementation of CompositeTensor is a relatively big feature. So it may not be implemented in the near future. Anyway, really thank you for your contribution. ^_^

Wanglongzhi2001:
LGTM @Oceania2018

Oceania2018 (approved):

PR #1184, opened 2023-09-27
test: Gradient descent  optimizer tests

novikov-alexander:


test/TensorFlowNET.UnitTest/PythonTest.cs
```diff
@@ -144,6 +144,37 @@ public void assertAllClose(double value, NDArray array2, double eps = 1e-5)
...
+                double delta = Math.Abs(a - b);
+                if (delta < _epsilon)
+                {
+                    return 0;
+                }
+                return a.CompareTo(b);
+            }
+        }
+
+        public void assertAllCloseAccordingToType<T>(
```
novikov-alexander, line 168: @Wanglongzhi2001 Does it make sense not to duplicate all assertion code but combine all in one assembly?

src/TensorFlowNET.Core/Variables/variables.py.cs
```diff
@@ -155,7 +157,10 @@ public static Operation _safe_initial_value_from_op(string name, Operation op, D
 
         public static Tensor global_variables_initializer()
         {
-            throw new NotImplementedException();
+            // if context.executing_eagerly():
+            //      return control_flow_ops.no_op(name = "global_variables_initializer")
+            var group =  variables_initializer(global_variables().ToArray());
```
novikov-alexander, line 162: @Wanglongzhi2001 an implicit cast happens there. `group` is actually an `Operation`. `Tensor` constructor goes to `Operation._ouputs.FirstOrDefault()` and returns `null` since `_outputs` for `no_op` are empty. This `null` than goes to `evaluate` functions and causes `NullReferenceException` since `null` can't be evaluated. 
So, the question is "shouldn't we return some placeholder instead of null"? If not than how should we rewrite the `evalute` function to support `no_op`?

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
```diff
@@ -0,0 +1,68 @@
...
+                var grads1 = tf.constant(new[] { 0.01, 0.01 }, dtype: dtype);
+                var optimizer = tf.train.GradientDescentOptimizer(3.0f);
+                var grads_and_vars = new[] {
+                    Tuple.Create(grads0, var0 as IVariableV1),
+                    Tuple.Create(grads1, var1 as IVariableV1)
+                };
+                var sgd_op = optimizer.apply_gradients(grads_and_vars);
+
+                var global_variables = variables.global_variables_initializer();
+                self.evaluate<T>(global_variables);
```
novikov-alexander, line 40: @Wanglongzhi2001 this `evaluate` fails because of reasons explained above.

src/TensorFlowNET.Core/Variables/variables.py.cs
```diff
@@ -155,7 +157,10 @@ public static Operation _safe_initial_value_from_op(string name, Operation op, D
 
         public static Tensor global_variables_initializer()
         {
-            throw new NotImplementedException();
```
Oceania2018, line 158: Should we use `tf.compat.v1.global_variables_initializer()`?

Oceania2018:
```console
 Failed TestBasic [1 s]
  Error Message:
   Test method Tensorflow.Keras.UnitTest.Optimizers.GradientDescentOptimizerTest.TestBasic threw exception: 
System.NullReferenceException: Object reference not set to an instance of an object.
  Stack Trace:
      at TensorFlowNET.UnitTest.PythonTest.evaluate[T](Tensor tensor) in D:\a\TensorFlow.NET\TensorFlow.NET\test\TensorFlowNET.UnitTest\PythonTest.cs:line 245
   at Tensorflow.Keras.UnitTest.Optimizers.GradientDescentOptimizerTest.TestBasicGeneric[T]() in D:\a\TensorFlow.NET\TensorFlow.NET\test\TensorFlowNET.UnitTest\Training\GradientDescentOptimizerTests.cs:line 40
   at Tensorflow.Keras.UnitTest.Optimizers.GradientDescentOptimizerTest.TestBasic() in D:\a\TensorFlow.NET\TensorFlow.NET\test\TensorFlowNET.UnitTest\Training\GradientDescentOptimizerTests.cs:line 62
```

src/TensorFlowNET.Core/Variables/variables.py.cs
Wanglongzhi2001: According to the implemention in tensorflow, we should return placeholder:
https://github.com/tensorflow/tensorflow/blob/6334682d8fcd1413eab2464e35696be48638a7a7/tensorflow/python/framework/test_util.py#L2801-L2810

src/TensorFlowNET.Core/Variables/variables.py.cs
novikov-alexander: I think so. Thanks!

src/TensorFlowNET.Core/Variables/variables.py.cs
novikov-alexander: I have the following exception when run on a placeholder Tensor: 
Tensorflow.InvalidArgumentError: 'You must feed a value for placeholder tensor 'Placeholder' with dtype float
	 [[{{node Placeholder}}]]'
	 
Should I choose a different dtype?

Oceania2018: Did you feed the placeholder with different type?

novikov-alexander: It turned out I just had to run `session.run(variable)` that initializes it.

src/TensorFlowNET.Core/Variables/variables.py.cs
novikov-alexander: Followed the advice

test/TensorFlowNET.UnitTest/PythonTest.cs
```diff
@@ -196,17 +245,25 @@ public T evaluate<T>(Tensor tensor)
             //    return self._eval_helper(tensors)
             //  else:
             {
-                var sess = tf.Session();
```
novikov-alexander, line 199: @Wanglongzhi2001 `ts.Session()` produces a new session where we need to get a cached one.

test/TensorFlowNET.UnitTest/PythonTest.cs
```diff
@@ -196,17 +245,25 @@ public T evaluate<T>(Tensor tensor)
...
+                var sess = tf.get_default_session();
```
novikov-alexander, line 248: @Wanglongzhi2001 The same functions exist in another assembly. 
Should we generalize them in one common assembly in another PR?

test/TensorFlowNET.UnitTest/PythonTest.cs
```diff
@@ -312,6 +450,54 @@ private Session _create_session(Graph graph, object cfg, bool forceGpu)
...
+                if (crash_if_inconsistent_args && self._cached_graph != null && !self._cached_graph.Equals(graph))
+                    throw new ValueError(@"The graph used to get the cached session is 
+                                           different than the one that was used to create the
+                                           session. Maybe create a new session with 
+                                           self.session()");
+                if (crash_if_inconsistent_args && self._cached_config != null && !self._cached_config.Equals(config))
+                {
+                    throw new ValueError(@"The config used to get the cached session is 
+                                           different than the one that was used to create the 
+                                           session. Maybe create a new session with
```
novikov-alexander, line 481: @Wanglongzhi2001 this code is also copy pasted from my previous PR because that's how test architecture implemented right now. Should I move it to one common assembly?

test/TensorFlowNET.UnitTest/PythonTest.cs
```diff
@@ -196,17 +245,25 @@ public T evaluate<T>(Tensor tensor)
             //    return self._eval_helper(tensors)
             //  else:
             {
-                var sess = tf.Session();
+                var sess = tf.get_default_session();
                 var ndarray = tensor.eval(sess);
```
novikov-alexander, line 249: @Wanglongzhi2001 what is the reason `tensor.eval` used there?
Maybe replace by `sess.run(tensor)` how it is made in original tests?

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+                // Run 1 step of sgd
+                sgd_op.run();
+                // Validate updated params
+                self.assertAllCloseAccordingToType(
+                    new[] { 1.0 - 3.0 * 0.1, 2.0 - 3.0 * 0.1 },
+                    self.evaluate<T[]>(var0));
+                self.assertAllCloseAccordingToType(
+                    new[] { 3.0 - 3.0 * 0.01, 4.0 - 3.0 * 0.01 },
+                    self.evaluate<T[]>(var1));
+                // TODO: self.assertEqual(0, len(optimizer.variables()));
```
novikov-alexander, line 61: @Wanglongzhi2001 Do I understand correctly that this one is not applicable for TensorFlow.NET?

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            // train.GradientDescentOptimizer is V1 only API.
+            tf.Graph().as_default();
+            using (var sess = self.cached_session())
+            {
+                var var0 = tf.Variable(new[] { 1.0, 2.0 }, dtype: dtype);
+                var var1 = tf.Variable(new[] { 3.0, 4.0 }, dtype: dtype);
+                var grads0 = tf.constant(new[] { 0.1, 0.1 }, dtype: dtype);
+                var grads1 = tf.constant(new[] { 0.01, 0.01 }, dtype: dtype);
+                var optimizer = tf.train.GradientDescentOptimizer(3.0f);
```
novikov-alexander, line 35: @Wanglongzhi2001 TensorFlow can accept lambda in this constructor and there's a test testing it.
Is it intentionally that TensorFlow.NET doesn't support that interface?

Wanglongzhi2001:
@novikov-alexander Please fix the ci error.

novikov-alexander:
> @novikov-alexander Please fix the ci error.

Not clear how they interfer. I'll try to debug.

Oceania2018 (approved):

novikov-alexander:
@Wanglongzhi2001 That's not clear.. I'm afraid that LinearRegression is flaky one. I just put my tests near and it caused some fail inside. But there are no visible reasons. Moreover it stopped to fail when I changed namespace of my tests.

novikov-alexander:
All checks have passed

Wanglongzhi2001:
@Oceania2018 LGTM.

test/TensorFlowNET.UnitTest/PythonTest.cs
AsakusaRinne: I think so, if you would like to.

test/TensorFlowNET.UnitTest/PythonTest.cs
AsakusaRinne: I think it's to map the tensor to a node in the session (graph).

test/TensorFlowNET.UnitTest/PythonTest.cs
AsakusaRinne: I think moving to a common namespace is a good idea if it could be re-used by other modules. Please open an another PR if you would like to do that.

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
AsakusaRinne: I don't think so, it's more like a compromise to the early fast-development. Therefore only a smallest and most-used implementation was provided. I think it's feasible to support it.

AsakusaRinne (approved):
Thank you a lot for this contribution! All things seems to be well done. If you'd like to change anything mentioned in the previous comments, please open a new PR. :)

PR #1192, opened 2023-10-10
fix: Implemented support for loading models with Concatenate layers

Jucko13:
This is my first serious pullrequest so please be gentle :)

The `tf.keras.models.load_model()` is now able to load models with this layer type.

There was a lot missing but I have a Unet model running in the field that now fully works.

I had some weird troubles with python tensorflow (im using 2.11.1) that savemodel would put the inbound nodes for concatenate layers in an extra nested array. My call in python looks like:
```python
comb = layers.Concatenate(axis=3, name=name+"_concat1")([concat_layer,conv])
```
This would create the following output in the manifest (notice the extra `[` and `]`:
```json
{
  "class_name": "Concatenate",
  "config": {
    "name": "EXP_1_concat1",
    "trainable": true,
    "dtype": "float32",
    "axis": 3
  },
  "name": "EXP_1_concat1",
  "inbound_nodes": [
    [
  	[
  	  "CNT_4_conv_2",
  	  0,
  	  0,
  	  {}
  	],
  	[
  	  "EXP_1_conv_0",
  	  0,
  	  0,
  	  {}
  	]
    ]
  ]
},
```
so thats why i added an extra check in generic_utils.cs for this extra nest. I would love some feedback on the implementation of this check and if it's the correct way to do.

And lastly, since concatenate layers have more than 1 input node we need a List<NodeConfig> when processing nodes in Functional.FromConfig.cs.

I also added a few tests to make sure that saving and loading a model with concatenate layers is the same (this was not the case before my changes).

### DIFF
--- a/src/TensorFlowNET.Core/Keras/ArgsDefinition/Merging/MergeArgs.cs
+++ b/src/TensorFlowNET.Core/Keras/ArgsDefinition/Merging/MergeArgs.cs
@@ -1,13 +1,15 @@
-﻿using System;
+﻿using Newtonsoft.Json;
+using System;
 using System.Collections.Generic;
 using System.Text;
 
 namespace Tensorflow.Keras.ArgsDefinition
 {
     // TODO: complete the implementation
-    public class MergeArgs : LayerArgs
+    public class MergeArgs : AutoSerializeLayerArgs
     {
         public Tensors Inputs { get; set; }
+        [JsonProperty("axis")]
         public int Axis { get; set; }
     }
 }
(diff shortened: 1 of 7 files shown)


Wanglongzhi2001:
Thank you very much for your contribution!

Wanglongzhi2001:
Looks good to me! @Oceania2018

Wanglongzhi2001:
> Do i have to do anything regarding the semantic check that failed? But to fix that I need to rename the commit title?
The reason for semantic checking error is the incorrect format of your commit title. The ci in TensorFlow.NET obey the angular style, You can change it to `fix: Implemented support for loading models with Concatenate layers`. But it's just a trivial question, you can discard it.

PR #1217, opened 2023-11-13
test: more gradient optimizer tests

novikov-alexander:


test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
```diff
@@ -67,6 +68,56 @@ public void TestBasic()
...
+        {
+            var dtype = GetTypeForNumericType<T>();
+
+            // train.GradientDescentOptimizer is V1 only API.
+            tf.Graph().as_default();
+            using (var sess = self.cached_session())
+            {
+                var var0 = tf.Variable(new[] { new[] { 1.0, 2.0 } }, dtype: dtype);
+                var var1 = tf.Variable(new[] { 3.0 }, dtype: dtype);
+                var x = tf.constant(
```
novikov-alexander, line 81: @Oceania2018 I can't initialize such a 2D constant. It either throws conversion exception or data allocation exception. Do I do anything wrong or there is a bug?

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
```diff
@@ -115,5 +166,72 @@ public void TestTensorLearningRate()
...
+                var var0 = tf.Variable(new[] { 1.0, 2.0 }, dtype: dtype);
+                var var1 = tf.Variable(new[] { 3.0, 4.0 }, dtype: dtype);
+                var grads0 = tf.constant(new[] { 0.1, 0.1 }, dtype: dtype);
+                var grads1 = tf.constant(new[] { 0.01, 0.01 }, dtype: dtype);
+                var grads_and_vars = new[] {
+                    Tuple.Create(grads0, var0 as IVariableV1),
+                    Tuple.Create(grads1, var1 as IVariableV1)
+                };
+                var sgd_op = tf.train.GradientDescentOptimizer(3.0f)
+                    .apply_gradients(grads_and_vars, global_step: global_step);
```
novikov-alexander, line 214: @AsakusaRinne why does `apply_gradiens` take `System.Tuple` while `zip` produces `(T1, T2)`. Would it be better to replace or extend `apply_gradients` interface to support valuetuple as well?

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
Wanglongzhi2001: Is it okay to float type? I'm not sure whether tf.net can initialize a 2D constant with double type, but it absolutly can when float type.

novikov-alexander: I just tried float type.
```
public static TensorProto make_tensor_proto(object values, TF_DataType dtype = TF_DataType.DtInvalid, Shape? shape = null, bool verify_shape = false, bool allow_broadcast = false)
{
   ...
   var len = dtype.get_datatype_size() * (int)shape.size;
   byte[] bytes = new byte[len];
   System.Buffer.BlockCopy(array, 0, bytes, 0, len);
```
`BlockCopy` expects that array contains only primitive values and throws there. 
Could you please give me an example where a float 2D constant works?

Wanglongzhi2001: Here is a example use float type:
https://github.com/SciSharp/TensorFlow.NET/blob/8cab730f98112a845461bb635f137e220f272adb/test/TensorFlowNET.Graph.UnitTest/ImageTest.cs#L54-L82

novikov-alexander: I see. It should be a 2D array `[,]` , not a jagged array `[][]`. Thanks.

test/TensorFlowNET.UnitTest/Training/GradientDescentOptimizerTests.cs
```diff
@@ -67,6 +68,51 @@ public void TestBasic()
...
+                sess.run(new[] { var0, var1 });
+                // Fetch params to validate initial values
+                self.assertAllCloseAccordingToType<T>(new[,] { { 1.0, 2.0 } }, self.evaluate<T[,]>(var0));
+                self.assertAllCloseAccordingToType(new[] { 3.0 }, self.evaluate<T[]>(var1));
+                // Run 1 step of sgd
+                sgd_op.run();
+                // Validate updated params
+                var np_pred = 1.0 * 4.0 + 2.0 * 5.0 + 3.0;
+                var np_grad = 2 * np_pred;
+                self.assertAllCloseAccordingToType(
```
novikov-alexander, line 99: @Wanglongzhi2001 Hm, now it calculates but the test doesn't pass. However, the code corresponds to TensorFlow original test. I have to check math there.

novikov-alexander: There just was a small typo, but I didn't have time to debug it :-D

novikov-alexander:
Hi @AsakusaRinne  @Wanglongzhi2001 
I fixed the mistake. Please review when you have bandwidth

PR #1234, opened 2024-02-12
fix: Avoid modifying augmented graph view collection upon traversal

barfeous:
Change code to avoid modifying the cache collection while maintaining the loops


AsakusaRinne (approved):
Thank you for the contribution!

Issue #1247, opened 2024-05-14
Deserialization exception thrown when loading a Keras SavedModel


SchoenTannenbaum:
### Description

Hi,
I recently started using TensorFlow.NET and was trying to load my saved Keras model, but I get an exception thrown during model deserialization.

Specifically:
1. I saved my original pretrained .h5 Keras model to a  .PB based SavedModel directory (following the instructions provided by @AsakusaRinne in [this closed issue](https://github.com/SciSharp/TensorFlow.NET/issues/1024) ) using a Python script running TF 2.15.0
1. I then tried to load it in C# using: ` var model = tf.keras.models.load_model("TestModel", false);`

However, I get this error: 
```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> Newtonsoft.Json.JsonSerializationException: Could not create an instance of type Tensorflow.Keras.IRegularizer. Type is an interface or abstract class and cannot be instantiated. Path 'bias_regularizer.class_name'.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader reader, JsonObjectContract objectContract, JsonProperty containerMember, JsonProperty containerProperty, String id, Boolean& createdFromNonDefaultCreator)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

PR #1248, opened 2024-05-16
Regularizer addition and fixes

SchoenTannenbaum:
Addresses load issue in : https://github.com/SciSharp/TensorFlow.NET/issues/1247


AsakusaRinne:
Hi, the overall looks good but could you please move your modification of the unit test to the following position? 

https://github.com/SciSharp/TensorFlow.NET/blob/master/test/TensorFlowNET.Keras.UnitTest/Model/ModelLoadTest.cs#L134

In this unit test, the model is saved and then loaded. Thus both the saving and loading for the regularizer could be tested.

